PR #5, opened 2020-01-10
Feature/add git hooks

mattch1:
Add precommit script to check feature flags and versions.


.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env sh
+exec < /dev/tty
+    while read -p "Have all the required feature flags been added to these changes? (y/n) " yn; do
+        case $yn in
+            [Yy] ) break;;
+            [Nn] ) echo "Please add any necessary feature flags!"; exit 1;;
```
mwestacott: Is the double ;; at end of line needed?

.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env sh
+exec < /dev/tty
```
mwestacott: Guessing that we'll add the 'npm run lint' command later?

.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
...
+        esac
+    done
+    while read -p "Is the version file up to date? (y/n) " yn; do
+        case $yn in
+            [Yy] ) break;;
+            [Nn] ) echo "If required please update the version file!"; exit 1;;
+            * ) echo "Please answer y (yes) or n (no):" && continue;
+        esac
+    done
+exec
```
mwestacott: Also wondering why extensions has '<&-' at the end of this line. Would it be needed here?

.githooks/pre-commit
mattch1: Yes that'll be when we add the node js code with the lint library.

.githooks/pre-commit
mattch1: In this code doule ;; are required.

PR #8, opened 2020-01-13
Lfa 1191 use typescript

mwestacott:


mattch1:
This is not using the app router yet, but this is something we'll look into as we introduce the pages.

PR #11, opened 2020-01-14
LFA-1194 Add initial screens

mwestacott:
Adds prototype content for the welcome and company number screens as well as introducing some new TS artefacts to mimic the set-up in extensions.

### DIFF
--- a/views/company-number.html
+++ b/views/company-number.html
@@ -7,13 +7,7 @@
   Tell us your details
 {% endblock %}
 
-{% block beforeContent %}
-  {{ govukPhaseBanner({
-    tag: {
-      text: "alpha"
-    },
-    html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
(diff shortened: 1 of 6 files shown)


views/company-number.html
```diff
@@ -0,0 +1,91 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  {% if errorList.length > 0 %}
+    Error:
+  {% endif %}
+  Tell us your details
+{% endblock %}
+
+{% block beforeContent %}
```
markpit: have a look at extensions-web, i think this might be in layout.html so it can be shared between pages.

views/index.html
```diff
@@ -1,26 +1,82 @@
 {% extends "layout.html" %}
 
 {% block pageTitle %}
-Promise to File - GOV.UK
+  Tell us a company is still required
```
SVarallo: Should be "Tell us if a company is still required"

PR #13, opened 2020-01-16
Lfa 1196 add unit test and config

mwestacott:


jest.config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  roots: [
+    "<rootDir>"
```
markpit: rootdir not defined?

test/routes/index.spec.unit.ts
```diff
@@ -0,0 +1,24 @@
+import app from "../../src/app";
+import * as request from "supertest";
+
+describe("Basic URL Tests", () => {
+
+  it("should find start page", async () => {
+    const response = await request(app)
+      .get("/promise-to-file");
+    expect(response.status).toEqual(200);
```
markpit: how do you know it's returning the correct page?  can you check page title or some identifying string in response.text (i think)

jest.config.js
markpit: ignore this, <rootDir> apparently points to the dir that contains package.json

mwestacott: Copied from SDK Node and extensions projects. Don't know what this does but will try removing it.

test/routes/index.spec.unit.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it("should find company number page", async () => {
+    const response = await request(app)
+      .get("/promise-to-file/company-number")
+      .set("Referer", "/")
```
markpit: Referer not used at this point. Might not be needed if we can find a way of capturing the page being requested via middleware for the back button.

mwestacott: Will remove 'Refererer' as not yet used or needed by this skeleton app.

PR #14, opened 2020-01-17
Added shell script to updatechs configs environments

mattch1:
Added a shell script to update chs configs environments for ptf web, the idea is to avoid manually having to go to each environment and update, instead we can update from a standard template.
A parameter is passed in "dev", "live" or "all".

A list is selected form the releveant file and added to a path for the environments selected, the ptf config is then updated in that enviroment using the template_env file which overwrites the content if the file exists.

If the file does not exist it creates the file with the content.

The use types in the shell file name with sh on the command line together with an argument "dev", "live" or "all".
The user is then prompted to ensure that they have checked out a branch in chs configs and to check that they have correctly selected dev/live/all.


markpit:
Is there a story for this?

mattch1:
It's been done as part of LFA 1192 there is an example PR for updating all the environment configs
https://github.com/companieshouse/chs-configs/pull/700/files so it relates to work that needs to be done. We can create a separate ticket if we need to.

environments/update_chs_configs.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
+
```
mwestacott: I think it's worth getting the detail on your initial PR comment into a comment block at the top of this shell script - explaining what it's for and how it's used. Probably also worth then referencing this script in the promise to file README.MD file (which I don't think exists yet).

environments/update_chs_configs.sh
```diff
@@ -0,0 +1,72 @@
...
+	     while read line
+	     do
+	        environments[$index]=$line
+	        index=$((index+1))
+	     done < dev	
+	     update environments
+	elif [ "$1" = "live" ]; then
+		 environments=(environments)
+	     index=0
+		 while read line
```
mwestacott: Indentation looks wrong on this line and the one in next block.

environments/update_chs_configs.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+echo "Ensure that you have a branch checked out from develop in chs configs"
+echo "Also you have selected <$1>, ensure that this correct"
+while true; do
+    read -p "Do you wish to continue?" yn
+    case $yn in
+        [Yy]* ) run $1; break;;
+        [Nn]* ) exit;;
+        * ) echo "Please answer yes or no.";;
```
mwestacott: Is it 'yes' or 'no' that needs entering or 'Y', 'y', 'N' or 'n'?

mattch1: y or n

environments/update_chs_configs.sh
mattch1: Added comment.

environments/update_chs_configs.sh
mattch1: Can be either.

environments/update_chs_configs.sh
mattch1: Indentations fixed

mattch1: For some reason the indentation looks perculiar in the diffs here on git but in the script file they are now lined up.

PR #15, opened 2020-01-17
pre-reqs for sonar pipeline integration

ICoxCH:
added npm scripts to run jes tests with coverage, and added sonar properties to allow publishing report to sonar


jest.config.coverage.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
mwestacott: As discussed, not sure we need this file as we have a similar one already pushed.

package.json
```diff
@@ -7,7 +7,9 @@
     "build": "tsc && cp -r views dist/",
     "devstart": "nodemon",
     "test": "jest --forceExit",
-    "lint": "tslint -p ."
+    "lint": "tslint -p .",
+    "sonarqube": "npm run coverage-jest && sonar-scanner",
+    "coverage-jest": "jest -c jest.config.coverage.js --forceExit --coverage"
```
mwestacott: Target could be just 'coverage', for consistency with the 'test' target above, which also uses Jest but doesn't mention it in the target name.

package.json
```diff
@@ -7,7 +7,9 @@
     "build": "tsc && cp -r views dist/",
     "devstart": "nodemon",
     "test": "jest --forceExit",
-    "lint": "tslint -p ."
+    "lint": "tslint -p .",
+    "sonarqube": "npm run coverage-jest && sonar-scanner",
+    "coverage-jest": "jest -c --forceExit --coverage"
```
mwestacott: See previous comment about the target name.

Also, just curious as to what the -c flag does?

ICoxCH: -c is if you want to define a non default config (I should of removed this when removing jest.config.coverage.ts)

PR #18, opened 2020-01-22
Add session handling

ICoxCH:
Copies the relevant code from extensions-web for session handling


src/services/redis.service.ts
```diff
@@ -0,0 +1,22 @@
+import Session from "../session/session";
+import redisStore from "../session/store/redis.store";
+
+/**
+ * Saves a session to redis.
+ * @param chSession
+ */
+const saveSession = async (chSession: Session): Promise<void> => {
+await redisStore.save(chSession);
```
mwestacott: Formatting looks wrong here (no indentation) and in the function below.

src/session/config.ts
```diff
@@ -0,0 +1,23 @@
+/**
```
mwestacott: How about naming this a bit more specifically environment.ts?

src/session/keys.ts
```diff
@@ -0,0 +1,20 @@
+export const SIGN_IN_INFO: string = "signin_info";
```
mwestacott: A comment at the top of the file would be useful to explain what these are, where they come from etc.

src/session/config.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Get's an environment variable. If the env var is not set and a default value is not
```
markpit: Get's should be Gets!

src/session/session.ts
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Static factory to create a new session instance using the default store
+     * (redis).
+     */
+    public static newInstance(): Session {
+        const session = new Session();
+        const now = Math.floor((new Date()).getTime() / 1000);
+        session.data = {
+            [keys.ID]: session.sessionKey(),
+            [keys.EXPIRES]: now + parseInt(DEFAULT_SESSION_EXPIRATION, 10),
```
mwestacott: This '10' is a magic number! Would be good to define a constant somewhere that describes what it actually is (number_of_minutes_before_session_expiry?) and reference it here.

src/session/keys.ts
```diff
@@ -0,0 +1,20 @@
...
+export const SIGNED_IN: string = "signed_in";
+export const ACCESS_TOKEN: string = "access_token";
+export const EXPIRES_IN: string = "expires_in";
+export const REFRESH_TOKEN: string = "refresh_token";
+export const ADMIN_PERMISSIONS: string = "admin_permissions";
+export const CLIENT_SIG: string = ".client.signature";
+export const EXPIRES: string = "expires";
+export const ID: string = ".id";
+export const LAST_ACCESS: string = "last_access";
```
markpit: some of these might not be needed, we can weed them out through the IDE

src/session/session.ts
```diff
@@ -0,0 +1,83 @@
...
+    private constructor(cookieId?: string) {
+        if (!cookieId) {
+            this.generateNewCookieId();
+        } else {
+            this._cookieId = cookieId;
+        }
+    }
+
+    public sessionKey(): string {
+        return this._cookieId.substr(0, 28);
```
mwestacott: Why do we only use the first 28 characters of the id? A comment here would be helpful.

src/model/error.messages.ts
```diff
@@ -0,0 +1 @@
+export const REDIS_PROBLEM: string = "There is a problem with redis: ";
```
mwestacott: Since there's only one entry in here (for now) wonder if this could just be defined in the 'redis.store.ts' file, or is it referenced elsewhere too?

src/session/store/util.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  let adminPermissions = false;
+  if (keys.ADMIN_PERMISSIONS in signInInfoData) {
+    adminPermissions = signInInfoData[keys.ADMIN_PERMISSIONS] === "1";
+  }
+
+  let signedIn = false;
+  if (keys.SIGNED_IN in signInInfoData) {
+    signedIn = signInInfoData[keys.SIGNED_IN] === 1;
```
mwestacott: Wondering why 1 above is a string but here it's a number. Again, what does the 1 signify and could it be defined as a constant that describes the value?

src/session/store/util.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @param data the entire blob of data from the store.
+ */
+export const unmarshalAccessToken = (data: IMap<any>): IAccessToken | undefined => {
+  const accessTokenData = data[keys.ACCESS_TOKEN];
+  if (!accessTokenData) {
+    return undefined;
+  }
+
+  return {
+    expiresIn: parseInt(accessTokenData[keys.EXPIRES_IN], 10),
```
mwestacott: There's that 10 again. We should defo define a constant for this I think.

src/session/keys.ts
markpit: ignore this, thought it was extensions-web stuff, not session code

test/routes/index.spec.unit.ts
```diff
@@ -1,29 +1,47 @@
...
+        return session;
+    });
+});
 
 describe("Basic URL Tests", () => {
 
-  it("should find start page", async () => {
-    const response = await request(app)
-      .get("/promise-to-file");
+    it("should find start page", async () => {
```
markpit: other files seem to be indented with 2 spaces

src/session/types.ts
```diff
@@ -0,0 +1,29 @@
+// arbitrary map type of string => T
+export interface IMap<T> {
+    [key: string]: T;
+}
+
+export interface IAccessToken {
+    token?: string;
```
mwestacott: Do the question marks actually do anything here? I thought they only come into play/become relevant when accessing sub-attributes or a potentially null object.

src/services/redis.service.ts
```diff
@@ -0,0 +1,22 @@
...
+const saveSession = async (chSession: Session): Promise<void> => {
+    await redisStore.save(chSession);
+};
+
+/**
+ * Loads a session from the redis store.
+ * @param cookieId
+ */
+const loadSession = async (cookieId: string): Promise<Session> => {
+    const session: Session = Session.newWithCookieId(cookieId);
```
markpit: 2 space indents

src/session/session.ts
```diff
@@ -0,0 +1,85 @@
+import {IMap} from "./types";
+import * as crypto from "crypto";
+import {COOKIE_SECRET, DEFAULT_SESSION_EXPIRATION, INTEGER_PARSE_BASE} from "../properties";
+import * as keys from "./keys";
+
+export default class Session {
+    get cookieId(): string {
```
markpit: indents

src/session/store/util.ts
mwestacott: Seems that ERIC returns differing data types so safer to work off the actual types that we receive.

PR #20, opened 2020-01-23
Added nunjucks template

mattch1:
Added nunjucks template 
To get it working added harcoded router values and a controller stub - these will be developed further on ticket lfa-1235

### DIFF
--- a/src/controllers/check.company.controller.ts
+++ b/src/controllers/check.company.controller.ts
@@ -1,0 +1,17 @@
+import { NextFunction, Request, Response } from "express";
+
+export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+
+    const scenario = { company: { PTFRequested: "Yes", accountsDue: "01 January 2020",
+            accountsStatus: "overdue", address: "25 No Street, Nowhere", companyType: "Private Limited Company",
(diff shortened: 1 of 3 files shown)


src/controllers/check.company.controller.ts
```diff
@@ -0,0 +1,17 @@
+import { NextFunction, Request, Response } from "express";
+
+export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+
```
mwestacott: I know it's obvious, but when I add code like this I tend to qualify it with a TODO comment, something like:

   // TODO Stub implementation to return dummy company details until they have been correctly retrieved

Up to you.

src/routes/routes.ts
```diff
@@ -16,5 +16,6 @@ const renderTemplate = (template: string) => (req: Request, res: Response, next:
 
 router.get(pageURLs.ROOT, renderTemplate(templatePaths.INDEX));
 router.get(pageURLs.COMPANY_NUMBER, renderTemplate(templatePaths.COMPANY_NUMBER));
+router.get("/check-company", checkCompanyRoute.route);
```
ICoxCH: Should this URL be in pageURLs as a constant?

mattch1: Yes it should - it will be worked on immediately afterwards on ticket lfa-1235

PR #22, opened 2020-01-23
Adding piwik environment variables to test setup

SVarallo:


test/global.setup.ts
```diff
@@ -6,5 +6,7 @@ export default async () => {
   process.env.COOKIE_NAME = "cookie_name";
   process.env.COOKIE_DOMAIN = "cookie_domain";
   process.env.COOKIE_SECURE_ONLY = "cookie_secure";
+  process.env.PIWIK_URL = "test";
```
ICoxCH: might be best if they were "piwik_url" and "piwik_site_id" just incase you need the values at any point (unlikely i know)

PR #23, opened 2020-01-23
Lfa 1222 implement sign in

mwestacott:


src/app.ts
```diff
@@ -22,30 +24,27 @@ const env = nunjucks.configure([
...
+app.use(sessionMiddleware);
+
+app.use(`${pageURLs.PROMISE_TO_FILE}/*`, authenticate);
 
 // view engine setup
 app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
 
-app.use(logger("dev"));
-app.use(express.json());
```
mwestacott: These 4 lines just needed to move up, above the authentication redirection config, for sign in to work.

src/app.ts
```diff
@@ -57,6 +56,9 @@ app.use((req, res, next) => {
 
 // error handler
 app.use((err, req, res, next) => {
+
+  logger.error("An error has occured. Re-routing to the error screen - " + err.stack);
```
mwestacott: In trying to get this working, I was getting redirected to the error page but nothing helpful was appearing in the logs. This new line ensures that the stack trace is written there when an error causes the redirection.

src/app.ts
```diff
@@ -22,30 +24,27 @@ const env = nunjucks.configure([
...
 // view engine setup
 app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
 
-app.use(logger("dev"));
-app.use(express.json());
-app.use(express.urlencoded({extended: false}));
-app.use(cookieParser());
-app.use(sessionMiddleware);
+app.use(morganLogger("dev"));
```
markpit: as discussed, look at combining morgan with log4js

src/authentication/middleware/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (!req.chSession.isSignedIn()) {
+
+    logger.debug("User not authenticated - redirecting to login screen");
+
+    let returnToUrl: string = pageURLs.PROMISE_TO_FILE;
+
+    const referringPageURL = req.header("Referer") as string;
+
+    // Redirect the user to to the start screen if they weren't already on the PTF journey
```
markpit: what happens if they are signed in but the referer is blank?

test/authentication/middleware/index.spec.unit.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it("should redirect to start screen if /promise-to-file/* called from outside of the PTF journey and not signed in", async () => {
+    setNotSignedIn();
+    const response = await request(app)
+      .get("/promise-to-file/company-number")
+      .expect("Location", "/promise-to-file");
+    expect(response.status).toEqual(302);
+  });
+
+
```
markpit: have you run tslint, i thought it moans about double blank lines.

test/routes/index.spec.unit.ts
```diff
@@ -2,35 +2,35 @@ import app from "../../src/app";
...
 import {loadMockSession} from "../mock.utils";
-import Session from "../../src/session/session";
+import {COOKIE_NAME} from "../../src/properties";
 
 jest.mock("../../src/services/redis.service");
 
 const mockCacheService = (<unknown>loadSession as jest.Mock<typeof loadSession>);
 
 beforeEach(() => {
+  mockCacheService.mockRestore();
```
markpit: mockRestore will replace the mock with it's original non-mocked implementation - https://jestjs.io/docs/en/mock-function-api#mockfnmockrestore.
maybe mockReset or mockClear might be more suitable.

src/app.ts
mwestacott: Yep. Will raise a new task under story LFA-1131 though and do this work under a separate PR.

test/authentication/middleware/index.spec.unit.ts
mwestacott: No. Will check.

test/routes/index.spec.unit.ts
mwestacott: I'll check. I did experiment with a couple of things here though and this was the only thing that worked. It's the same code that we had in extensions...

src/app.ts
mwestacott: https://companieshouse.atlassian.net/browse/LFA-1271

test/routes/index.spec.unit.ts
mwestacott: Turns out the mockRestore() call isn't required at all, so removing this line.

test/authentication/middleware/index.spec.unit.ts
mwestacott: Right, tslint config looks slightly wrong because it's only picking up violations on the 'src' folder and not 'test'. We also want to ensure that test JS resources don't go in the 'dist' folder. @markpit to raise a new tech story to cover this work.

src/authentication/middleware/index.ts
markpit: As discussed, it's unlikely to be a problem.

mwestacott: Seems that user is allowed to jump into the journey at that point. It's potentially something we want to address but this is the same behaviour as Extensions, so probably not urgent.

PR #24, opened 2020-01-23
Company number screen validation

SVarallo:
This covers the basic validation for the company number screen including "no company number", "invalid company number" and "company number too long".

### DIFF
--- a/src/controllers/company.number.controller.ts
+++ b/src/controllers/company.number.controller.ts
@@ -5,7 +5,7 @@ import {createGovUkErrorData, GovUkErrorData} from "../model/govuk.error.data";
 import * as templatePaths from "../model/template.paths";
 import {ValidationError} from "../model/validation.error";
 
-// validator middleware
+// validator middleware that checks for an empty or too long input
 const preValidators = [
   check("companyNumber").blacklist(" ").escape().not().isEmpty().withMessage(errorMessages.NO_COMPANY_NUMBER_SUPPLIED),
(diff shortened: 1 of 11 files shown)


src/controllers/company.number.controller.ts
```diff
@@ -0,0 +1,57 @@
+import {NextFunction, Request, Response} from "express";
+import {check, validationResult} from "express-validator/check";
+import * as errorMessages from "../model/error.messages";
+import {createGovUkErrorData, GovUkErrorData} from "../model/govuk.error.data";
+import * as templatePaths from "../model/template.paths";
+import {ValidationError} from "../model/validation.error";
+
+// validator middleware
```
mwestacott: '// validator middleware' here and on the function below isn't a very descriptive comment. Could you add some more detail, like you did for 'padCompanyNumber' below?

test/controllers/company.number.controller.spec.unit.ts
```diff
@@ -0,0 +1,86 @@
...
+
+jest.mock("../../src/services/redis.service");
+
+const NO_COMPANY_NUMBER_SUPPLIED = "No company number supplied";
+const INVALID_COMPANY_NUMBER = "Invalid company number";
+const COMPANY_NUMBER_TOO_LONG = "Company number too long";
+
+describe("company number validation tests", () => {
+
+   // This will be needed when the routes are secured
```
mwestacott: Personal preference maybe, but I like to use a tag (something like TODO) in the comment, just so this can be searched for later on to find any blocks of code like this that need to be revisited/reincluded.

PR #25, opened 2020-01-27
LFA-1271 Correctly configure the Morgan HTTP logger

mwestacott:
Also integrate it with Log4js

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,7 +1,8 @@
 import { configure, getLogger } from "log4js";
+import { LOG_LEVEL } from "./properties";
 
 const logger = getLogger("promise-to-file-web");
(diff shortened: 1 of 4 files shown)


src/logger.ts
```diff
@@ -1,7 +1,8 @@
 import { configure, getLogger } from "log4js";
+import { LOG_LEVEL } from "./properties";
 
 const logger = getLogger("promise-to-file-web");
-logger.level = process.env.LOG_LEVEL || "info";
+logger.level = LOG_LEVEL;
```
mwestacott: Corrected, as this was never right. Environment properties should come through the 'properties.ts' file and default values are set there too.

PR #26, opened 2020-01-27
Feature/lfa 1235 router and controller for check company

mattch1:
Added:
Code to handle the company number to the company number controller.

Api client to obtain company profile data from the company profile api.
Api enumerations.
Session service.

The check company controller code to render the company profile api data for that comapny number.
Updated the check company html to take a boolean rather than an "overdue" string and to check whether the date itself is overdue.

Unit tests for check copmpany

TO DO:
Get confirmation statement and ptf data from company profile method of the sdk .
Test the new confirmation statment fields once sdk obtains this data from company profile api.

Create overdue page and create ptf request if not overdue.
Test accounts overdue page when this is in place

Place the company profile sdk call somewhere more global so that it not called in the check company as well as the company number controller.

Investigate API Enumerations setup.

Complete session handling including the code around session creation.


src/client/api.enumerations.ts
```diff
@@ -0,0 +1,16 @@
+import * as fs from "fs";
+import * as yaml from "js-yaml";
+
```
markpit: indentation should be 2 spaces

src/model/page.urls.ts
```diff
@@ -1,5 +1,8 @@
 export const ROOT: string = "/";
+export const PTF: string = "/promise-to-file";
```
SVarallo: Isn't this already covered by `PROMISE_TO_FILE` below?

src/client/date.formatter.ts
```diff
@@ -0,0 +1,12 @@
+import moment = require("moment");
+
+const DISPLAY_DATE_FORMAT: string = "D MMMM YYYY";
+const REASON_DATE_FORMAT: string = "YYYY-MM-DD";
+
+export const formatDateForDisplay = (inputDate: string): string => {
+    return moment(inputDate).format(DISPLAY_DATE_FORMAT);
+};
+
+export const formatDateForReason = (day: string, month: string, year: string): string => {
```
markpit: PTF won'r have reason, this function is not needed

src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,45 @@
...
-            accountsStatus: "overdue", address: "25 No Street, Nowhere", companyType: "Private Limited Company",
-            csDue: "02 January 2020", csStatus: "overdue", incorporationDate: "01 January 2000",
-            name: "Test Company LTD", number: "00001111", status: "Active" }};
+    const companyNumber: string = sessionService.getCompanyInContext(req.chSession);
+    if (companyNumber) {
+        try {
+            logger.info(`Company number ${companyNumber} found in session, retrieving company profile`);
+            const token: string = req.chSession.accessToken() as string;
+            const companyinContext: PTFCompanyProfile = await getCompanyProfile(companyNumber, token);
+            logger.info("COMPANY PROFILE " + JSON.stringify(companyinContext));
```
markpit: I think it would be better to log out the company profile only in debug.

src/controllers/company.number.controller.ts
```diff
@@ -52,6 +57,39 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
+
+  const companyNumber: string = req.body.companyNumber;
+  try {
+    logger.info(`Retrieving company profile for company number ${companyNumber}`);
+    const token: string = req.chSession.accessToken() as string;
+
+    // TODO: Place this call somewhere more global so that it not called on check company as well
+    const company: PTFCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+    await sessionService.createPromiseToFileSession(req.chSession, company.companyNumber);
```
markpit: add a check so that you only create a session if one doesn't exist

src/session/config.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Get's an environment variable. If the env var is not set and a default value is not
+ * provided, then it is assumed it is a mandatory requirement and an error will be
+ * thrown.
+ */
+const getEnvironmentValue = (key: string, defaultValue?: any): string => {
```
markpit: this file is not needed, see properties.ts

src/services/session.service.ts
```diff
@@ -0,0 +1,32 @@
+import Session from "../session/session";
```
markpit: This file should be covered by unit tests as well

src/client/date.formatter.ts
```diff
@@ -0,0 +1,12 @@
+import moment = require("moment");
```
markpit: missing unit tests

src/controllers/check.company.controller.ts
mwestacott: I've changed the config so that debug output appears in dev and test environments now too (you may need to merge in 'master' for this to work).

src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,45 @@
...
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
 
 export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 
-    const scenario = { company: { PTFRequested: "Yes", accountsDue: "01 January 2020",
-            accountsStatus: "overdue", address: "25 No Street, Nowhere", companyType: "Private Limited Company",
-            csDue: "02 January 2020", csStatus: "overdue", incorporationDate: "01 January 2000",
-            name: "Test Company LTD", number: "00001111", status: "Active" }};
+    const companyNumber: string = sessionService.getCompanyInContext(req.chSession);
```
mwestacott: This looks a bit odd - if a company is in context and stored in the session then why do we need to go and retrieve it again? I'm guessing that 'getCompanyInContext' on the session object is wrong and it should be named 'getCompanyNumberInContext'?

src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,45 @@
...
+            const token: string = req.chSession.accessToken() as string;
+            const companyinContext: PTFCompanyProfile = await getCompanyProfile(companyNumber, token);
+            logger.info("COMPANY PROFILE " + JSON.stringify(companyinContext));
+            const isDueDatePassed: boolean = checkDueDate(companyinContext);
+            return res.render(templatePaths.CHECK_COMPANY, {
+                    company: companyinContext,
+                    dueDatePassed: isDueDatePassed,
+                });
+        } catch (e) {
+            logger.error(`Error retrieving company number ${companyNumber} from redis`, e);
```
mwestacott: Could this error message be mis-leading - what if there is actually a company number in redis but the call to the SDK on line 16 fails? This catch block would trigger (presumably) and message would be wrong. Perhaps just make the error message more generic.

src/model/company.profile.ts
```diff
@@ -0,0 +1,19 @@
...
+    address: {
+        line_1: string;
+        line_2: string;
+        postCode: string;
+    };
+    accountsDue: string;
+    accountingPeriodStartOn: string;
+    accountingPeriodEndOn: string;
+    isAccountsOverdue: boolean;
+    isCSOverdue: boolean;
```
mwestacott: Would suggest not shortening ConfirmationStatement to CS, but up to you.

src/model/page.urls.ts
```diff
@@ -1,5 +1,8 @@
...
 export const PROMISE_TO_FILE: string = "/promise-to-file";
 export const COMPANY_NUMBER: string = "/company-number";
+export const CHECK_COMPANY: string = "/check-company";
+export const PTF_CHECK_COMPANY: string = PTF + CHECK_COMPANY;
```
mwestacott: We should standardise on shortening 'promise to file' to 'ptf' in variable/constant names if this is preferable (so refactor the constant name below at this point too).

src/services/session.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    await saveSession(chSession);
+    return ptfSession;
+};
+
+/**
+ * Returns the company in context. That is the most recent company that was
+ * input in the company number screen.
+ * @param chSession
+ */
+const getCompanyInContext = (chSession: Session): string => {
```
mwestacott: Yes, so rename this to be 'getCompanyNumberInContext' since it holds a company number and not a company.

src/session/types.ts
```diff
@@ -27,3 +27,14 @@ export interface ISignInInfo {
...
 }
+
+export interface IPromiseToFileSession {
+  company_in_context: string;
+  ptf_requests: IPromiseToFileRequest[];
+}
+
+export interface IPromiseToFileRequest {
+  company_number: string;
+  extension_request_id: string;
```
mwestacott: Guessing this shouldn't be named 'extension_request_id' but 'promise_to_file_request_id'?

src/model/page.urls.ts
mattch1: Using the abbeviation "PTF" for long names but for the URL itself we use dthe full name "PROMISE_TO_FILE" this way we get to be as brief as possible here we need to be and as descriptiive as possible where can be.

src/model/company.profile.ts
mattch1: Godd idea I'll make this more descriptive

src/controllers/check.company.controller.ts
mattch1: Shorted so that it doe snot mention redis just the error retrieving the company.

src/controllers/check.company.controller.ts
mattch1: Yes it's confusing - will change this to say number

src/session/config.ts
mattch1: Removed

src/client/date.formatter.ts
mattch1: deleted

src/model/page.urls.ts
mwestacott: Makes sense.

views/check-company.html
```diff
@@ -21,48 +21,57 @@
...
+
+    {% set address %}
+        {{company.address.line_1}}
+        <br/>
+        {{company.address.line_2}}
+        <br/>
+        {{company.address.postCode}}
+    {% endset %}
+
     {{ govukTable({
```
SVarallo: This should be a govukSummaryList element and not a table. While used in the prototype this was deemed not fit for purpose by the DAC report and was subsequently changed to a summary list instead. Have talked to Ollie and he agrees on this. See link below or extensions for the layout.

https://design-system.service.gov.uk/components/summary-list/

src/controllers/check.company.controller.ts
mattch1: Removed this as it is not needed.

src/controllers/company.number.controller.ts
mattch1: Added the check to the create function itself.

src/services/session.service.ts
mattch1: changed

src/session/types.ts
mattch1: changed

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+        company_number_in_context: companyNumber,
+        ptf_requests: [],
+    };
+    chSession.appendData(keys.PTF_SESSION, ptfSession);
+    await saveSession(chSession);
+    return ptfSession;
+};
+
+/**
+ * Returns the company in context. That is the most recent company that was
```
mwestacott: Comment needs updating since it's a number and not a company that's returned.

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+    chSession: Session, companyNumber: string): Promise<IPromiseToFileSession> => {
+
+    const existing = chSession.data[keys.PTF_SESSION];
+    if (existing) {
+        return existing;
+    }
+
+    const ptfSession: IPromiseToFileSession = {
+        company_number_in_context: companyNumber,
+        ptf_requests: [],
```
mwestacott: Do we need this yet? Probably not, since this PR is all about retrieving company details based on the number.

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+    return ptfSession;
+};
+
+/**
+ * Returns the company in context. That is the most recent company that was
+ * input in the company number screen.
+ * @param chSession
+ */
+const getCompanyNumberInContext = (chSession: Session): string => {
+    return chSession.data.ptf_session.company_in_context;
```
mwestacott: COMPANY_IN_CONTEXT is in the keys.ts file - should this be used here instead? Also, why is it getting a company and not a number (or is this another naming issue?).

src/model/company.profile.ts
```diff
@@ -0,0 +1,19 @@
...
+        postCode: string;
+    };
+    accountsDue: string;
+    accountingPeriodStartOn: string;
+    accountingPeriodEndOn: string;
+    isAccountsOverdue: boolean;
+    isConfirmationStatementOverdue: boolean;
+    confirmationStatementDue: string;
+    incorporationDate: string;
+    ptfRequested: string;
```
mwestacott: Wondering if we need this now? What is it for?

mattch1: These will be required as we will need to display information about confirmation stements alongside accounts - there work to do to get this data from the sdk.

src/controllers/company.number.controller.ts
mattch1: The check added was causing problems so removing this.

mattch1: As we are not doing this in extensions-web and so will need investigating so I suggest we do this on another ticket.

mattch1: Added update function which can be called from the company number controller, removed check from create function.

markpit: can you move the create session to a middleware function to be called from app.ts after the authorisation? that way all routes will be covered and not just this one.

src/services/session.service.ts
mattch1: Unit test added.

src/client/date.formatter.ts
mattch1: Unit test added.

src/services/session.service.ts
mattch1: Removed

src/services/session.service.ts
mattch1: Name changed

src/services/session.service.ts
mattch1: Chanegd.

src/controllers/company.number.controller.ts
```diff
@@ -52,6 +58,45 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
+
+    // TODO: Place this call somewhere more global so that it not called on check company as well
+    const company: PTFCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+    const existing = req.chSession.data[keys.PTF_SESSION];
+    if (existing) {
+      await sessionService.updatePTFSessionValue(req.chSession,
+          keys.COMPANY_NUMBER_IN_CONTEXT, company.companyNumber);
+    } else  {
+      await sessionService.createPromiseToFileSession(req.chSession, company.companyNumber);
```
markpit: It would be better if this was created by middleware for all routes.

src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,45 @@
...
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
 
 export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 
-    const scenario = { company: { PTFRequested: "Yes", accountsDue: "01 January 2020",
-            accountsStatus: "overdue", address: "25 No Street, Nowhere", companyType: "Private Limited Company",
-            csDue: "02 January 2020", csStatus: "overdue", incorporationDate: "01 January 2000",
-            name: "Test Company LTD", number: "00001111", status: "Active" }};
+    const companyNumber: string = sessionService.getCompanyNumberInContext(req.chSession);
```
markpit: indentation should be 2 spaces

src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,45 @@
...
+            const token: string = req.chSession.accessToken() as string;
+            const companyinContext: PTFCompanyProfile = await getCompanyProfile(companyNumber, token);
+            logger.debug("COMPANY PROFILE " + JSON.stringify(companyinContext));
+            const isDueDatePassed: boolean = checkDueDate(companyinContext);
+            return res.render(templatePaths.CHECK_COMPANY, {
+                    company: companyinContext,
+                    dueDatePassed: isDueDatePassed,
+                });
+        } catch (e) {
+            logger.error(`Error retrieving company number ${companyNumber}`, e);
```
mwestacott: It's an error retrieving the company, not the company number. Would recommend updating the log message.

src/controllers/company.number.controller.ts
```diff
@@ -52,6 +58,45 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
+    }
+
+    return res.redirect(pageURLs.PTF_CHECK_COMPANY);
+  } catch (e) {
+    logger.error(`Error fetching company profile for company number ${companyNumber}`, e);
+    if (e.status === 404) {
+      buildError(res, errorMessages.COMPANY_NUMBER_NOT_FOUND);
```
mwestacott: Looks like the wrong error message key is being used here - it's the company that's not found and not the company number.

src/model/error.messages.ts
```diff
@@ -1,5 +1,6 @@
 export const ERROR_SUMMARY_TITLE: string = "There is a problem";
-
 export const INVALID_COMPANY_NUMBER: string = "Invalid company number";
 export const NO_COMPANY_NUMBER_SUPPLIED: string = "No company number supplied";
 export const COMPANY_NUMBER_TOO_LONG: string = "Company number too long";
+export const NO_COMPANY_NUMBER_IN_SESSION: string = "No company number found in session";
+export const COMPANY_NUMBER_NOT_FOUND: string = "Company number not found";
```
mwestacott: See previous comment - should this be 'Company not found'?

src/controllers/company.number.controller.ts
mattch1: I agree but I would like to focus on this as part fo another ticket as I think there are few issues to look at with session handling so will add a TODO here.

src/session/keys.ts
```diff
@@ -20,3 +20,7 @@ export const LOCALE: string = "locale";
 export const PERMISSIONS: string = "permissions";
 export const SCOPE: string = "scope";
 export const COMPANY_NUMBER: string = "company_number";
+export const PTF_SESSION: string = "ptf_session";
+export const PTF_REQUESTS: string = "ptf_requests";
```
mwestacott: I don't think we need PTF_REQUESTS (yet).

src/controllers/check.company.controller.ts
mattch1: How about "Error retrieving company with number <company number>"?

src/controllers/company.number.controller.ts
mattch1: This was done as part of ticket LFA-1136 but will update this.

src/model/error.messages.ts
mattch1: Yes it should.

src/controllers/check.company.controller.ts
mattch1: Updated formatting.

src/session/keys.ts
mattch1: Removed.

views/check-company.html
```diff
@@ -21,113 +21,120 @@
...
+                    html: CSoverdueHTML
+                }
             }
         ]
-        ]
     }) }}
 
-    {% if scenario.company.PTFRequested !== 'yes' %}
+    {% if company.ptfRequested !== 'yes' %}
     {{ govukButton({
```
SVarallo: This will need ```
attributes: {
          "data-event-id": "continue-button",
          id: "submit"
        }
```

File: src/controllers/check.company.controller.ts (hunk 22 above)
Comment (mwestacott) [reply src/controllers/check.company.controller.ts]: Yep, sounds good.

File: .gitmodules [hunk 27]
```diff
@@ -0,0 +1,3 @@
+[submodule "api-enumerations"]
```
ICoxCH: The Makefile is going to need to be updated to ensure the submodule is present when building the service.
Adding a new update_submodules target to the makefile such as https://github.com/companieshouse/extensions-web/blob/master/Makefile#L67

and calling it as a pre-req for the build target such as: https://github.com/companieshouse/extensions-web/blob/master/Makefile#L21

Should ensure this

mattch1:
Makefile will be updated on another ticket as there were problem with the command which requires investigation

mwestacott:
@mattch1 can this PR now be closed? I think everything it was trying to solve has been addressed by other tasks and merged branches.

mattch1:
Closing this review as the work has been split up and completed on other reviews.

PR #27, opened 2020-01-27
disable ptf session creation

ICoxCH:
Disabled session creation and setting a cookie from within the ptf service as this was causing authentication errors, the session and cookie will be created when the user is directed to the sign in service

### DIFF
--- a/src/authentication/middleware/index.ts
+++ b/src/authentication/middleware/index.ts
@@ -8,6 +8,8 @@ export default (req: Request, res: Response, next: NextFunction) => {
 
   if (!req.chSession.isSignedIn()) {
 
+    logger.debug("User not authenticated");
+
     let returnToUrl: string = pageURLs.PROMISE_TO_FILE;
 
     const referringPageURL = req.header("Referer") as string;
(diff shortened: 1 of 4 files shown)


src/authentication/middleware/index.ts
```diff
@@ -9,21 +8,21 @@ export default (req: Request, res: Response, next: NextFunction) => {
 
   if (!req.chSession.isSignedIn()) {
```
mwestacott: I would still have a 'User not authenticated' message in the log.

src/session/middleware/index.ts
```diff
@@ -15,18 +14,13 @@ declare global {
...
-    session.setClientSignature(req.get("user-agent") || "", req.ip);
-    await redisService.saveSession(session);
 
-    req.chSession = session;
-    res.cookie(COOKIE_NAME, session.cookieId, {path: "/"});
-  } else {
+  if (cookieId) {
     logger.info("cookie found, loading session from redis: " + cookieId);
     req.chSession = await redisService.loadSession(cookieId);
+  }else {
```
mwestacott: A space required before the 'else' I think. Did you run ts-lint? Thought this would pick it up...

markpit:
Can you talk to Matt re: session creation. His PR https://github.com/companieshouse/promise-to-file-web/pull/26/files is creating a new session in the company number controller

PR #28, opened 2020-01-27
LFA-1273 Configure ts-lint so that test code is also analysed

mwestacott:


package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc && cp -r views dist/",
     "devstart": "nodemon",
     "test": "jest --forceExit",
-    "lint": "tslint -p .",
+    "lint": "tslint '{src,test}/**/*'",
```
mwestacott: Specify the directories to analyse rather than use the settings in the tsconfig.json file (which deliberately doesn't write generated test JS files to the 'dist' directory).

tslint.json
```diff
@@ -11,7 +11,6 @@
   },
   "linterOptions": {
     "exclude": [
-      "src/test/**/*.ts"
```
mwestacott: Not required - there is no src/test dir.

PR #30, opened 2020-01-29
LFA-1215 Redirect to sign-in screen if not authenticated

mwestacott:


test/authentication/middleware/index.spec.unit.ts
```diff
@@ -60,7 +60,7 @@ describe("Authentication middleware", () => {
     setNotSignedIn();
     const response = await request(app)
       .get("/promise-to-file/company-number")
-      .expect("Location", "/promise-to-file");
+      .expect("Location", "/signin?return_to=/promise-to-file");
```
mattch1: Not sure why "/promise-to-file" is appended here.

PR #33, opened 2020-01-30
Updated template and route for check company page

SVarallo:


src/controllers/check.company.controller.ts
```diff
@@ -1,17 +1,17 @@
...
+  const scenario = { company: { PTFRequested: "Yes", accountsDue: "01 January 2020",
+    accountsStatus: "overdue", address: "25 No Street, Nowhere, nl2br", type: "Private Limited Company",
+    csDue: "02 January 2020", csStatus: "overdue", incorporationDate: "01 January 2000",
+    name: "Test Company LTD", number: "00001111", status: "Active" }};
 
-    return res.render("check-company", {
-        scenario,
-    });
+  return res.render(templatePaths.CHECK_COMPANY, {
+    scenario,
```
markpit: We should remove references to scenario as they were used by the UX team to populate their prototype screens ie. test scenarios. So pass company instead and in the template scenario.company.stuff would become company.stuff.

package.json
```diff
@@ -17,7 +17,7 @@
     "debug": "~2.6.9",
     "express": "~4.16.1",
     "express-validator": "^5.3.1",
-    "govuk-frontend": "^3.2.0",
+    "govuk-frontend": "^3.5.0",
```
SVarallo: This is at Ollie's recommendation

PR #35, opened 2020-01-30
Feature/lfa 1333 session mw call

mattch1:
Renamed existing session middleware to "cookie middelware" and put it under a cookie directory and added a new mddileware index directly under the "middleware" dir  to create the session.

Added a unit test to test the create methods in session service.

Dev tested the middleware to check that in company number we can refer to the session update it with a company number without an undefined error being thrown. Result was as expected with the company number being returned in the logs.


src/app.ts
```diff
@@ -36,6 +37,7 @@ app.enable("trust proxy");
 app.use(express.json());
 app.use(express.urlencoded({ extended: false }));
 app.use(cookieParser());
+app.use(cookieMiddleware);
```
markpit: Looking at cookieMiddleware it basically loads the chSession. I think it would be more descriptive to name it chSessionLoader instead of cookieMiddleware.

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Creates a new ptf session and adds a company to the context.
+ * This will override any ptf session that is already in redis.
+ * @param chSession
+ * @param companyNumber
+ */
+const createPromiseToFileSession = async (
+  chSession: Session): Promise<IPromiseToFileSession> => {
+  const ptfSession: IPromiseToFileSession = {
+    company_number: "",
```
mwestacott: When we create the new session, do we / should we add an empty company number value? Why bother? This could just be added when a client requests to add it to the PTF session. Alternatively, when we call this function, we have to supply a company number as a parameter.

src/app.ts
markpit: or set it back to sessionMiddleware and call the new middleware 'ptfSessionLoader'

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+  chSession.appendData(keys.PTF_SESSION, ptfSession);
+  await saveSession(chSession);
+  return ptfSession;
+};
+
+const updatePromiseToFileSessionValue = async (chSession: Session, key: string, value: any): Promise<void> => {
```
mwestacott: Rather than have the 'generic' update method that takes a key, why not have specific update methods for company number, company details etc. This would then match the specific 'get' method immediately below. Alternatively, could make the get method below more generic by passing in a key. This interface currently uses a mixture of the two approaches, which is not ideal.

src/services/session.service.ts
mattch1: We may not have a company number at this point and to have the ptf session type we need at least one field.

src/services/session.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  const ptfSession = await chSession.data.ptf_session;
```
markpit: use keys.PTF_SESSION here

src/session/session.ts
```diff
@@ -76,6 +76,17 @@ export default class Session {
     this.data[keys.CLIENT_SIG] = data.digest("hex");
   }
 
+  public appendData(key: string, value: any): void {
+    this.data[key] = value;
+  }
+
+  public accessToken(): string | undefined {
```
mwestacott: I can't see where this method is called. Is it needed (right now)?

src/session/types.ts
```diff
@@ -27,3 +27,7 @@ export interface ISignInInfo {
   signedIn?: boolean;
   userProfile?: IUserProfile;
 }
+
+export interface IPromiseToFileSession {
+  company_number: string;
```
mwestacott: Again, we've got that COMPANY_NUMBER constant in keys.ts. Should it be used/referenced here or should it perhaps be deleted?

src/app.ts
markpit: Actually, we don't need to rename sessionMiddleware. Keep it as it was and just add the new middleware called something like ptfSessionLoader

src/services/session.service.ts
mwestacott: Suggests that the create method should be parameterised then with a company number. An empty session essentially serves no value.

src/app.ts
mattch1: I have come up with something else see app ts. I thinks we should have some naming symmetry between the two middlewares.

src/session/session.ts
mattch1: called from SessionService

test/services/session.service.spec.unit.ts
```diff
@@ -0,0 +1,13 @@
...
+import * as keys from "../../src/session/keys";
+
+describe ("cache service tests", () => {
+
+  it("should create a new empty promise to file session", async () => {
+    const session: Session = Session.newInstance();
+    await sessionService.createPromiseToFileSession(session);
+    await sessionService.updatePromiseToFileSessionValue(session, keys.COMPANY_NUMBER, "00006400");
+    expect(session.data.ptf_session.company_number).toEqual("00006400");
+  });
```
markpit: needs a test to check it doesn't create a new ptf session if one exists

src/session/types.ts
mattch1: It should be renamed.

src/services/session.service.ts
mattch1: Using this.

src/app.ts
mattch1: Put original name of cookie middleware back and descriptively named ptf session middleware.

src/services/session.service.ts
mattch1: Parameterised.

src/services/session.service.ts
mattch1: Using more generic get method, now used in test.

test/services/session.service.spec.unit.ts
mattch1: Testing the midleware here is tricky as it involves explicitly getting the mocked ch session. 

Attempted to write a unit test earlier but there currently seems to be no handle on this ch session. The code does not seemed to be designed for testing this directly. 

Suggest covering this further down the line when we have completed the comapny number and check company pages along the lines of what is in the index.spec.unit tests for the authentication mdidleware - these provide a more indirect way of testing middleware.

PR #37, opened 2020-01-30
Joining Page Urls and Template Paths

markpit:
Importing Template Paths into Page Urls to avoid duplications (there were a lot of dups in extensions).
Also moved a page url only used by test into the test.

### DIFF
--- a/src/model/page.urls.ts
+++ b/src/model/page.urls.ts
@@ -1,5 +1,10 @@
-export const ROOT: string = "/";
-export const PROMISE_TO_FILE: string = "/promise-to-file";
-export const COMPANY_NUMBER: string = "/company-number";
+import * as TEMPLATE from "./template.paths";
 
-export const PROMISE_TO_FILE_COMPANY_NUMBER: string = PROMISE_TO_FILE + COMPANY_NUMBER;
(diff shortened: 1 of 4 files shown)


src/model/page.urls.ts
```diff
@@ -1,5 +1,10 @@
...
+
+/**
+ * Keep template names in template.paths.ts and reference them in here for URLs
+ */
 
-export const PROMISE_TO_FILE_COMPANY_NUMBER: string = PROMISE_TO_FILE + COMPANY_NUMBER;
+const SEPARATOR: string = "/";
+export const ROOT: string = "/";
+export const PROMISE_TO_FILE: string = SEPARATOR + "promise-to-file";
+export const COMPANY_NUMBER: string = SEPARATOR + TEMPLATE.COMPANY_NUMBER;
```
mwestacott: This company number page URL looks wrong - there's no preceding '/promise-to-file' prefix.

markpit: Added some comments to the file to explain that the urls extend a base url set in the app.

markpit:
I won't merge this yet, need to tweak it a bit

PR #40, opened 2020-02-03
Changing page.urls to use template.paths names

markpit:
To avoid duplication of page names, I've changed page.urls to use the template names from template.paths.

### DIFF
--- a/src/model/page.urls.ts
+++ b/src/model/page.urls.ts
@@ -1,7 +1,14 @@
+import * as template from "./template.paths";
+
+/**
+ * Keep template names in template.paths.ts and reference them in here for URLs
+ */
+
 export const ROOT: string = "/";
-export const PROMISE_TO_FILE: string = "/promise-to-file";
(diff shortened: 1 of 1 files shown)


src/model/page.urls.ts
```diff
@@ -1,7 +1,14 @@
+import * as template from "./template.paths";
+
+/**
+ * Keep template names in template.paths.ts and reference them in here for URLs
```
mwestacott: Think it's worth expanding on the comment to explain why we have both COMPANY_NUMBER and PROMISE_TO_FILE_COMPANY_NUMBER constants (and others). I know we discussed this and understood why but I've forgotten now. Can you remember why @markpit ?

markpit: PROMISE_TO_FILE_XXX is probably used for redirects

markpit: comments updated

mwestacott: Perhaps say these are used for 'render' calls?

PR #43, opened 2020-02-05
Updated text on check company screen to match story and removed unneeded div from layout

SVarallo:


views/layout.html
```diff
@@ -35,8 +35,6 @@
   <script src="/html5-shiv/html5shiv.js"></script>
   <![endif]-->
 
-  <div id="templateName" data-id='{{templateName}}' hidden></div>
```
mattch1: Do we know what this was originally supposed to do?

PR #44, opened 2020-02-05
LFA-1373 Correctly set PTF overdue flags using dates returned

mwestacott:
I took a slightly different approach to extensions when implementing this - setting the correct boolean flag value in the PTF version of the company profile after retrieving and mapping it. It keeps complexity out of the controller and view and means the correct value is available throughout the PTF journey without having to (potentially) recalculate it.


test/client/api.client.spec.unit.ts
```diff
@@ -35,6 +35,49 @@ describe("apiclient company profile unit tests", () => {
...
     expect.assertions(1);
   });
+
+  it("correctly sets the overdue accounts flag if false in company profile but due date has passed",
+      async () => {
+    // Both here and in the tests below, create a deep copy of the dummy company profile before changing
+    // values, to ensure test isolation
+    const companyWithOverriddenAccountsNextDueDate: Resource<CompanyProfile> =
+        JSON.parse(JSON.stringify(dummySDKResponse));
+    companyWithOverriddenAccountsNextDueDate.resource.accounts.nextDue = "2019-05-31";
```
mattch1: Are you setting a fixed "current" date to test these dates against int he tests or just against the system time?

mwestacott: No, but just seen that the extensions test does something like this to prevent tests which pass now from failing in the future, so will address this.

mwestacott: Now setting the 'current' date, so this comment should now be addressed.

mwestacott:
Also added some further boundary testing around accounts due yesterday and today, now that I'm controlling today's date.

PR #45, opened 2020-02-06
Add company authentication

ICoxCH:


package.json
```diff
@@ -23,6 +23,7 @@
     "morgan": "~1.9.1",
     "moment": "^2.24.0",
     "msgpack5": "^4.2.1",
+    "node-jose": "^1.1.3",
```
mwestacott: Just curious as to what this new library adds/does?

ICoxCH: This library is for encrypting and signing a jwt, which we need for sending the 'state' paramater to the account service for authentication. 
This library is written by cisco, and recommended by jwt.io for a js implementation of jwt encryption.

src/app.ts
```diff
@@ -10,6 +10,7 @@ import sessionMiddleware from "./session/middleware";
 import ptfSessionLoader from "./session/middleware/ptf.session";
 import {PIWIK_SITE_ID, PIWIK_URL} from "./properties";
 import authenticate from "./authentication/middleware/index";
+import companyAuthenticate from "./web.security/middleware/index";
```
mwestacott: Just wondering about the package structure here. Should we maybe have './authentication/user/middleware' for the user login stuff (would mean a minor refactoring) and then place this new company authentication alongside in './authentication/company/middleware'?

src/web.security/jwt.encryption.ts
```diff
@@ -0,0 +1,42 @@
+import * as crypto from "crypto";
+import {OAUTH2_REQUEST_KEY} from "../properties";
+import * as jose from "node-jose";
+
```
mwestacott: I think this implementation might be ported from the Java equivalent version somewhere. Worth referencing (the repo?) in a general comment at the top of this file I think.

src/web.security/jwt.encryption.ts
```diff
@@ -0,0 +1,42 @@
+import * as crypto from "crypto";
```
mwestacott: If the 'web.security' package does remain, does this name follow/violate a naming standard? Just curious as I haven't seen any other packages with .s in.

src/web.security/middleware/index.ts
```diff
@@ -0,0 +1,80 @@
+import {NextFunction, Request, Response} from "express";
+import {COOKIE_NAME, OAUTH2_AUTH_URI, OAUTH2_CLIENT_ID,
+  OAUTH2_REDIRECT_URI} from "../../properties";
+import * as redisService from "../../services/redis.service";
+import * as keys from "../../session/keys";
+import logger from "../../logger";
+import {jweEncodeWithNonce, generateNonce} from "../jwt.encryption";
+
+const OAUTH_SCOPE_PREFIX = "https://api.companieshouse.gov.uk/company/";
+const COMPANY_NUMBER_KEY = "company_number";
```
mwestacott: This is also newly defined in keys.ts above. Are they essentially the same thing and could you just import/reference that one here instead?

src/web.security/middleware/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    req.chSession = await redisService.loadSession(cookieId);
+    const signInInfo = req.chSession.getSignedInInfo();
+    if (isAuthorisedForCompany(signInInfo, companyNumber)) {
+        logger.info("User is authenticated for %s", companyNumber);
+        return next();
+    } else {
+      return res.redirect(await getAuthRedirectUri(req, companyNumber));
+    }
+  } else {
+    // TODO: Process no session
```
mwestacott: Is there a JIRA task or story for this outstanding functionality? Would be good to create and reference it here.

src/web.security/middleware/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    return next(new Error("No session present for company auth filter"));
+  }
+};
+
+function getCompanyFromPath(path: string): string {
```
mwestacott: Would rename this function 'getCompanyNumberFromPath'.

src/web.security/middleware/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  const session = req.chSession;
+  const nonce = generateNonce();
+  session.data[keys.NONCE] = nonce;
+  await redisService.saveSession(session);
+
+  return await createAuthUri(originalUrl, nonce, scope);
+}
+
+async function createAuthUri(originalUri: string, nonce: string, scope?: string): Promise<string> {
+  let authUri = "";
```
markpit: these 2 lines can be merged - let authURI = authUri.concat...

test/web.security/jwt.encryption.spec.unit.ts
```diff
@@ -0,0 +1,46 @@
+import * as jwt from "../../src/web.security/jwt.encryption";
+import {OAUTH2_REQUEST_KEY} from "../../src/properties";
+import * as jose from "node-jose";
+import {generateNonce} from "../../src/web.security/jwt.encryption";
+
+jest.mock("../../src/session/store/redis.store", () => {
```
markpit: You might not have merged master in - you can use this mock syntax instead 
jest.mock("../../src/session/store/redis.store", () => import("../mocks/redis.store.mock.factory"));

views/continue-trading.html
```diff
@@ -0,0 +1,30 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Tell us if a company is still required
+{% endblock %}
+
+{% block header %}
+  <!-- Blank header with no service name for the start page -->
```
markpit: this references the start page - is it needed?

views/continue-trading.html
```diff
@@ -0,0 +1,30 @@
...
+      <li class="govuk-breadcrumbs__list-item">
+        <a class="govuk-breadcrumbs__link" href="https://www.gov.uk/browse/business/limited-company">Running a limited company</a>
+      </li>
+    </ol>
+  </div>
+{% endblock %}
+
+{% block content %}
+  <h1>CONTINUE TRADING PAGE</h1>
+  {% endblock %}
```
markpit: alignment

views/continue-trading.html
```diff
@@ -0,0 +1,29 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Tell us if a company is still required
+{% endblock %}
+
+{% block header %}
```
markpit: blockHeader is in layout.html - does this page require a different header to the one in the layout?

markpit: this is a dummy page until story for this is picked up.

PR #50, opened 2020-02-10
Added validation to still required page

SVarallo:
Also updated survey link to match that shown here: https://companieshouse.atlassian.net/browse/LFA-1165

### DIFF
--- a/src/controllers/still.required.controller.ts
+++ b/src/controllers/still.required.controller.ts
@@ -1,17 +1,58 @@
 import {NextFunction, Request, Response} from "express";
-import {PTFCompanyProfile} from "../model/company.profile";
+import {check, validationResult} from "express-validator/check";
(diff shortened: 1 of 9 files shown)


src/controllers/still.required.controller.ts
```diff
@@ -1,17 +1,58 @@
...
+
+  // render errors in the view
+  if (!errors.isEmpty()) {
+    const errorText = errors.array()
+      .map((err: ValidationError) => err.msg)
+      .pop() as string;
+
+    return renderPageWithError(res, errorText, companyName);
+  } else {
+    return res.render(STILL_REQUIRED, {
```
markpit: should this go to the next page after still required?

SVarallo: Yes it will but was leaving that open for now

PR #51, opened 2020-02-10
Fixing code smells highlighted by Sonar

markpit:
Sonar highlighted some code smells so I've fixed them based on the recommended Sonar fixes.

### DIFF
--- a/src/bin/www.ts
+++ b/src/bin/www.ts
@@ -28,7 +28,6 @@ const server = http.createServer(app);
 
 server.listen(port);
 server.on("error", onError);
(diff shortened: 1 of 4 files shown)


src/bin/www.ts
```diff
@@ -77,14 +76,3 @@ function onError(error) {
       throw error;
   }
 }
-
-/**
- * Event listener for HTTP server "listening" event.
- */
-
-function onListening() {
-  const addr = server.address();
```
markpit: these vars weren't being used, so we are left with an empty function. I've tested the app and it works without this.

PR #53, opened 2020-02-10
template for no longer required

mattch1:
The template for lfa-1364 counterpart can be found on PR: 
https://github.com/companieshouse/promise-to-file-web/pull/54

### DIFF
--- a/views/confirmation-not-required.html
+++ b/views/confirmation-not-required.html
@@ -2,9 +2,9 @@
 
 {% block pageTitle %}
     {% if errorList.length > 0 %}
-        Error:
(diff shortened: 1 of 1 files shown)


views/confirmation-not-required.html
```diff
@@ -0,0 +1,43 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+{% if errorList.length > 0 %}
```
markpit: this block indentation is missing

mattch1: Fixed this.

PR #55, opened 2020-02-11
Feature/lfa 1365 confirmation controller

mattch1:
The controller and routing for the confirmation pages.


src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const ACCOUNTS_NEW_DEADLINE: number = 28;
+const CS_NEW_DEADLINE: number = 14;
+
+const createMissingError = (item: string): Error => {
+    const errMsg: string = item + " missing from session";
+    return new Error(errMsg);
+};
+
+const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
markpit: needs unit tests

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
+import { NextFunction, Request, Response } from "express";
+import {formatDateForDisplay} from "../client/date.formatter";
+import logger from "../logger";
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
+import {ALREADY_SUBMITTED, COMPANY_NUMBER, COMPANY_PROFILE, SIGN_IN_INFO, USER_PROFILE} from "../session/keys";
+import {IUserProfile} from "../session/types";
+
+const ACCOUNTS_NEW_DEADLINE: number = 28;
```
mwestacott: I can't see anything in the story https://companieshouse.atlassian.net/browse/LFA-1166 that relates to the numbers 14 and 28. What are these? Code should relate to a business requirement somewhere.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+  const companyProfile: PTFCompanyProfile =
+      await sessionService.getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE);
+  const signInInfo = req.chSession.getSignedInInfo();
+  const userProfile: IUserProfile = signInInfo[USER_PROFILE] as IUserProfile;
+
+  const email = userProfile.email;
+  const dateRequested: Date = new Date(Date.now());
+  const extensionPeriodInDays =
+      (companyProfile.isAccountsOverdue) ? ACCOUNTS_NEW_DEADLINE : CS_NEW_DEADLINE;
```
mwestacott: If those constants remain in this PR I'd recommend renaming to closer match the variable name here - 'extensionPeriodInDays'. e.g. ACCOUNTS_EXTENSION_PERIOD_IN_DAYS (if that's what these are).

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
+import {ALREADY_SUBMITTED, COMPANY_NUMBER, COMPANY_PROFILE, SIGN_IN_INFO, USER_PROFILE} from "../session/keys";
+import {IUserProfile} from "../session/types";
+
+const ACCOUNTS_NEW_DEADLINE: number = 28;
+const CS_NEW_DEADLINE: number = 14;
+
+const createMissingError = (item: string): Error => {
```
markpit: this function is unused

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  const extension: Date = new Date(dateRequested);
+  extension.setDate(dateRequested.getDate() + extensionPeriodInDays);
+
+  const isSubmitted: boolean = await sessionService.getPromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED);
+  const token: string = req.chSession.accessToken() as string;
+
+  if (!(isSubmitted && token)) {
+    try {
+      await sessionService.updatePromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED, true);
+      // TODO LFA-TBC call promise-to-file api
```
mwestacott: Would be good to find a JIRA ticket id for this.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+    } catch (e) {
+       logger.error("Error processing application " + JSON.stringify(e));
+       await sessionService.updatePromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED, false);
+       return next(e);
+    }
+  } else {
+    logger.error("Form already submitted, not processing again");
```
mwestacott: Probably worth logging a bit more info here - 'form' is a bit generic. Which form?

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  const companyProfile: PTFCompanyProfile =
+      await sessionService.getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE);
+  const signInInfo = req.chSession.getSignedInInfo();
+  const userProfile: IUserProfile = signInInfo[USER_PROFILE] as IUserProfile;
+
+  const email = userProfile.email;
+  const dateRequested: Date = new Date(Date.now());
+  const extensionPeriodInDays =
+      (companyProfile.isAccountsOverdue) ? ACCOUNTS_NEW_DEADLINE : CS_NEW_DEADLINE;
+  const extension: Date = new Date(dateRequested);
```
mwestacott: Would this 'extension' variable be better named 'newDeadline'?

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+import {formatDateForDisplay} from "../client/date.formatter";
+import logger from "../logger";
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
+import {ALREADY_SUBMITTED, COMPANY_NUMBER, COMPANY_PROFILE, SIGN_IN_INFO, USER_PROFILE} from "../session/keys";
+import {IUserProfile} from "../session/types";
+
+const ACCOUNTS_NEW_DEADLINE: number = 28;
+const CS_NEW_DEADLINE: number = 14;
```
ICoxCH: Could these have more descriptive variable names? Such as "CONFIRMATION_STATEMENT_EXTENSION_DAYS"
Make it clear the value is the number of days granted for that type of extension

src/controllers/confirmation.controller.ts
ICoxCH: If neither accounts or confirmation statement are overdue, is this still the correct behaviour?

src/controllers/confirmation.controller.ts
ICoxCH: Is it possible to just construct this date to the correct value in the first place?

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  const dateRequested: Date = new Date(Date.now());
+  const extensionPeriodInDays =
+      (companyProfile.isAccountsOverdue) ? ACCOUNTS_NEW_DEADLINE : CS_NEW_DEADLINE;
+  const extension: Date = new Date(dateRequested);
+  extension.setDate(dateRequested.getDate() + extensionPeriodInDays);
+
+  const isSubmitted: boolean = await sessionService.getPromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED);
+  const token: string = req.chSession.accessToken() as string;
+
+  if (!(isSubmitted && token)) {
```
ICoxCH: Is this logic correct? Would it not be `!isSubmitted && token`?

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const email = userProfile.email;
+  const dateRequested: Date = new Date(Date.now());
+  const extensionPeriodInDays =
+      (companyProfile.isAccountsOverdue) ? ACCOUNTS_NEW_DEADLINE : CS_NEW_DEADLINE;
+  const extension: Date = new Date(dateRequested);
+  extension.setDate(dateRequested.getDate() + extensionPeriodInDays);
+
+  const isSubmitted: boolean = await sessionService.getPromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED);
+  const token: string = req.chSession.accessToken() as string;
```
markpit: don't need token as it is never used in this function

markpit: sorry, i can see the todo for api, ignore this comment

src/controllers/confirmation.controller.ts
mattch1: I'd prefer to avoid using the word "extension" in this project

src/controllers/confirmation.controller.ts
mattch1: I believe the extra ptf period is taken from the current date.

src/controllers/confirmation.controller.ts
mattch1: I'll removed for now but we will need it later on when we need to use the api for CHIPS contact and emails.

src/controllers/confirmation.controller.ts
markpit: i think the if check should just be on isSubmitted. We don't need token until we call the api so the get token can go inside the if and it will only be called when we need it. Also, if submitted = false and token is false then it will attempt to call the api but fail as token is false.

views/layout.html
```diff
@@ -18,6 +18,7 @@
 {% from "govuk/components/textarea/macro.njk"      import govukTextarea %}
 {% from "govuk/components/file-upload/macro.njk"   import govukFileUpload %}
 {% from "govuk/components/inset-text/macro.njk"    import govukInsetText %}
+{% from "govuk/components/warning-text/macro.njk"  import govukWarningText %}
```
markpit: is this needed? I don't see any nunjucks changes using this in this pull request.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+  const companyProfile: PTFCompanyProfile =
+      await sessionService.getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE);
+  const signInInfo = req.chSession.getSignedInInfo();
+  const userProfile: IUserProfile = signInInfo[USER_PROFILE] as IUserProfile;
+
+  const email = userProfile.email;
+  const dateRequested: Date = new Date(Date.now());
+  const extensionPeriodInDays =
```
markpit: probably best not to use the word extension in this project

mwestacott: So what is this period of time if not an extension? A stay of execution? additionalNumberOfDaysGranted?

views/layout.html
mattch1: confirmation-still-required.html  on line 52

mattch1: It should have gone on with ticket LFA 1364

src/controllers/confirmation.controller.ts
mattch1: Addressed for comment 2 above

src/controllers/confirmation.controller.ts
mattch1: This highlights an interesting point - presumably we are not going to get to this point if neither are overdue an alternative page should be displayed after clicking continue on the check company page stating this.

src/controllers/confirmation.controller.ts
mattch1: In line with my comment above about "extensions" these'll be changed.

src/controllers/confirmation.controller.ts
mattch1: Made the commit this complicated: Removed both submitted and token checks for now as they are not needed immediately will not add anything that cannot be tested first so as to avoid any mistakes.

mattch1: These should not be here until we are using the api.

src/controllers/confirmation.controller.ts
mattch1: Strictly speak no it is not related to 1166 but there is only one ticket for the controller and this will cover both the no longer and is still required alternatives. Given this we will need these values.

src/controllers/confirmation.controller.ts
mattch1: Will call it updated deadline.

src/controllers/confirmation.controller.ts
mattch1: added unit tests

src/controllers/confirmation.controller.ts
mattch1: It is used now.

src/controllers/confirmation.controller.ts
mattch1: Will avoid the name "extension" - all been renamed.

src/controllers/confirmation.controller.ts
mattch1: It would.

src/controllers/confirmation.controller.ts
mattch1: Mention now that it is a promise to file from

src/controllers/confirmation.controller.ts
mattch1: Have added something to deal with this scenareo.

src/controllers/confirmation.controller.ts
mwestacott: But where's the evidence that it's 14 not 15 days? Your word against mine or developers' word against testers. I really think that we should request an analyst to at least add an addendum to the story in JIRA or a short comment clarifying this.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,65 @@
+import { NextFunction, Request, Response } from "express";
+import {formatDateForDisplay} from "../client/date.formatter";
+import logger from "../logger";
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
+import {ALREADY_SUBMITTED, COMPANY_NUMBER, COMPANY_PROFILE, SIGN_IN_INFO, USER_PROFILE} from "../session/keys";
+import {IUserProfile} from "../session/types";
+
+const ACCOUNTS_EXT_DEADLINE_IN_DAYS: number = 28;
```
mwestacott: I still think that we should ask an analyst to update the story with the number of days. Without it, we have zero traceability to requirements.

mattch1: These are really just placeholders for the time being; they can still be changed.

src/controllers/confirmation.controller.ts
mattch1: This controller will change as we add to the project along with the TODO this is really just something to be going on with so that we can start the work for other tickets but we need to do so in order to bring many of thes changes about. This is in no way written in stone.

src/controllers/confirmation.controller.ts
mwestacott: Ok but what I'm struggling with is that there's no requirements/story around any of this logic to extend the deadline, including the number of days applied - at least it's not in the story that this PR relates to. So, should this code even be in this PR if not relevant now/yet? I suppose you could clarify it by adding a TODO comment to the code, but then what JIRA ticket will this hang off?

mattch1: I can add a TODO that's the best idea. As for which jira ticket - this returns me to the point that there is only one ticket to cover the controller - I will reference case LFA-1169 which deals with the confirmation for company is still reqired as this is what it relates to.

mwestacott: Ok, thanks. At least we then have some kind of reference and it's a reminder to check this later.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,65 @@
+import { NextFunction, Request, Response } from "express";
+import {formatDateForDisplay} from "../client/date.formatter";
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
+import * as sessionService from "../services/session.service";
+import {ALREADY_SUBMITTED, COMPANY_NUMBER, COMPANY_PROFILE, SIGN_IN_INFO, USER_PROFILE} from "../session/keys";
```
markpit: ALREADY_SUBMITTED, COMPANY_NUMBER and SIGN_IN_INFO . are unused

test/controllers/confirmation.controller.spec.unit.ts
```diff
@@ -0,0 +1,76 @@
...
+const EMAIL: string = "test@demo.ch.gov.uk";
+const PAGE_TITLE: string = "Confirmation page";
+const URL: string = pageURLs.PROMISE_TO_FILE + "/company/" + COMPANY_NUMBER + "/confirmation";
+const ERROR_PAGE: string = "Sorry, there is a problem with the service";
+
+describe("confirmation screen stating that the company is no longer required", () => {
+
+  const mockCacheService = loadSession as jest.Mock;
+  const mockPTFSession =  getPromiseToFileSessionValue as jest.Mock;
+
```
markpit: lint is failing due to trailing whitespace. Can you run 'make githooks' to add lint to the commit hook.

mattch1: Removed whitespace

src/controllers/confirmation.controller.ts
mattch1: Gone

src/controllers/confirmation.controller.ts
ICoxCH: I would tend to agree with Martin, in lfa-1365, or it's parent story, lfa-1166. There's no story for any logic in this controller barring retrieving company name from the session, probably best to not include any logic until there is a story for it, rather than placeholder logic we risk leaving in that's not founded in user requirements.
But we should decide this as a team

mattch1: Alright I'll remove this.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,50 @@
+import { NextFunction, Request, Response } from "express";
+import {PTFCompanyProfile} from "../model/company.profile";
+import * as templatePaths from "../model/template.paths";
```
SVarallo: You only need 2 variables here so there's no need to import all of them

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  const signInInfo = req.chSession.getSignedInInfo();
+  const userProfile: IUserProfile = signInInfo[USER_PROFILE] as IUserProfile;
+  const email = userProfile.email;
+
+  if (!email) {
+    return next(createMissingError("User Email"));
+  }
+  // TODO LFA-1169 Calrify what the ext deadline period will be for accounts and cs.
```
SVarallo: Typo on clarify

test/mock.utils.ts
```diff
@@ -24,21 +24,36 @@ export const loadMockSession = (mockLoadSessionFunction: jest.Mock<typeof loadSe
 };
 
 export const loadCompanyAuthenticatedSession = (mockLoadSessionFunction: jest.Mock<typeof loadSession>,
-                                                companyNumber?: string): void => {
+                                                companyNumber, email): void => {
```
markpit: you could make email optional with email?, then you wouldn't need to pass "" as email in some tests.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+  // TODO LFA-1380 need to add decision "is company still required" as a flag to the session.
+  const isRequired: boolean = false;
+  if (isRequired) {
+    return res.render(CONFIRMATION_STILL_REQUIRED,
+     {
+       company: companyProfile,
+       userEmail: email,
+     });
+  } else {
+    return res.render(CONFIRMATION_NOT_REQUIRED,
```
mwestacott: Sorry, should have spotted this earlier, but shouldn't this constant be 'COMPANY_NOT_REQUIRED' rather than 'CONFIRMATION...'?

Also, wondering why there's code added for the 'still required' scenario because that's not the screen related to this story - https://companieshouse.atlassian.net/browse/LFA-1166. Guessing that's a later requirement/story?

src/model/template.paths.ts
```diff
@@ -2,3 +2,5 @@ export const INDEX: string = "index";
 export const COMPANY_NUMBER: string = "company-number";
 export const CHECK_COMPANY: string = "check-company";
 export const STILL_REQUIRED: string = "still-required";
+export const CONFIRMATION_STILL_REQUIRED: string = "confirmation-still-required";
```
mwestacott: So yes, see comment above. I think these constants should refer to 'company' and not 'confirmation'.

src/controllers/confirmation.controller.ts
mattch1: No it refers to the comfirmation screen we need soemthingt hat corresponds to the url name.

src/model/template.paths.ts
mattch1: They need to refer to 'confirmation'

src/controllers/confirmation.controller.ts
mwestacott: Ah I see (company is kind of implied here). 👍

mwestacott: Ok, understand about the company/confirmation now but I still wonder why this PR is doing stuff related to the 'still required' scenario?

test/mock.utils.ts
```diff
@@ -24,21 +24,36 @@ export const loadMockSession = (mockLoadSessionFunction: jest.Mock<typeof loadSe
 };
 
 export const loadCompanyAuthenticatedSession = (mockLoadSessionFunction: jest.Mock<typeof loadSession>,
-                                                companyNumber?: string): void => {
+                                                companyNumber?, email?): void => {
```
ICoxCH: should this be `companyNumber?: string, email?: string`

test/controllers/confirmation.controller.spec.unit.ts
```diff
@@ -0,0 +1,76 @@
+import * as request from "supertest";
+import app from "../../src/app";
+import * as pageURLs from "../../src/model/page.urls";
```
ICoxCH: This file only needs to import the PROMISE_TO_FILE from page.urls

test/mock.utils.ts
mattch1: We need these to be null for some tests.

markpit: think he meant missing :string types

test/controllers/confirmation.controller.spec.unit.ts
mattch1: Minimsed the imports.

mattch1:
This could be to do with the way we have split the tickets up, we should consider making parent tickets scope related - so at present all want are confirmation pages that **display** information and what that information is. We have split still required and no longer required across different parent tickets. 

In hindsight it would have been better to have a parent that is scoped to display only and then  a subtask for each template then a controller subtask and then a navigation subtask. Case LFA-1320 could have then been given a different scope that builds on what we do here.

test/mock.utils.ts
mattch1: I know - typing was creating issues here - but I think I have found a way around this.

PR #57, opened 2020-02-14
Feature/lfa 1380 navigation to confirmation

mattch1:
Decision handling and routing to confirmation page - if yes a flag of true is added to the session and passed to the confirmation page, false otherwise.


src/controllers/still.required.controller.ts
```diff
@@ -34,13 +36,18 @@ const postRoute = async (req: Request, res: Response, next: NextFunction): Promi
 
     return renderPageWithError(res, errorText, companyName);
   } else {
-    return res.render(STILL_REQUIRED, {
-      companyName,
-      templateName: STILL_REQUIRED,
-    });
+    await addDecisionToSession(req.body.stillRequired, req.chSession);
+    const companyNumber = getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE).companyNumber;
```
markpit: you're getting the company profile from the session twice in this function. I would get the company profile once and store it in a const then use the const to get company number and company name from it.

src/controllers/still.required.controller.ts
```diff
@@ -34,13 +36,18 @@ const postRoute = async (req: Request, res: Response, next: NextFunction): Promi
...
+    const url = PROMISE_TO_FILE + "/company/" + companyNumber + APPEND_CONFIRMATION;
+    return res.redirect(url);
   }
 };
 
+const addDecisionToSession = async (decision: string, session: Session): Promise<void> => {
+  const decisionFlag: boolean = decision === "yes";
```
markpit: I would convert decision to uppercase and check ==="YES", that way this function won't break if the template changes to Yes or YEs or yes etc..

src/model/page.urls.ts
```diff
@@ -16,6 +16,7 @@ export const COMPANY_AUTH_PROTECTED_ROUTE: string = SEPARATOR + "company/:compan
  */
 export const STILL_REQUIRED: string = COMPANY_AUTH_PROTECTED_ROUTE + template.STILL_REQUIRED;
 export const CONFIRMATION: string = COMPANY_AUTH_PROTECTED_ROUTE + "confirmation";
+export const APPEND_CONFIRMATION: string = SEPARATOR + "confirmation";
```
markpit: make "confirmation" a const to avoid duplication

src/controllers/still.required.controller.ts
```diff
@@ -34,13 +36,18 @@ const postRoute = async (req: Request, res: Response, next: NextFunction): Promi
 
     return renderPageWithError(res, errorText, companyName);
   } else {
-    return res.render(STILL_REQUIRED, {
-      companyName,
-      templateName: STILL_REQUIRED,
-    });
+    await addDecisionToSession(req.body.stillRequired, req.chSession);
+    const companyNumber = getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE).companyNumber;
+    const url = PROMISE_TO_FILE + "/company/" + companyNumber + APPEND_CONFIRMATION;
```
ICoxCH: Could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
To replace :companyNumber in the route, with the company number we want, would save declaring the APPEND_CONFIRMATION constant.
Something like `const url = CONFIRMATION.replace(":companyNumber", companyNumber")`

test/controllers/still.required.controllers.spec.unit.ts
```diff
@@ -34,4 +33,26 @@ describe("still required validation tests", () => {
...
+
+  it("should update session when yes is selected", async () => {
+    const session: Session = Session.newInstance();
+    const response = await request(app)
+        .post("/promise-to-file/company/00006400/still-required")
+        .set("Referer", "/")
+        .set("Cookie", [`${COOKIE_NAME}=123`])
+        .send({stillRequired: true});
+
+    expect(response.status).toEqual(302);
```
markpit: missing the session value check

mattch1: I know - not sure how to check this in a jest unit test we are mocking the session.

test/controllers/still.required.controllers.spec.unit.ts
```diff
@@ -34,4 +33,26 @@ describe("still required validation tests", () => {
...
+
+  it("should update session when no is selected", async () => {
+
+    const response = await request(app)
+        .post("/promise-to-file/company/00006400/still-required")
+        .set("Referer", "/")
+        .set("Cookie", [`${COOKIE_NAME}=123`])
+        .send({stillRequired: false});
+
+    expect(response.status).toEqual(302);
```
markpit: missing the session value check

test/controllers/still.required.controllers.spec.unit.ts
markpit: ok i'll have a think

src/model/page.urls.ts
```diff
@@ -10,14 +10,18 @@ export const PROMISE_TO_FILE: string = SEPARATOR + "promise-to-file";
 export const COMPANY_NUMBER: string = SEPARATOR + template.COMPANY_NUMBER;
 export const CHECK_COMPANY: string = SEPARATOR + template.CHECK_COMPANY;
 export const COMPANY_AUTH_PROTECTED_ROUTE: string = SEPARATOR + "company/:companyNumber/";
+export const CONFIRMATION_APPEND: string = "confirmation";
 
 /**
  * Company Auth protected routes
  */
 export const STILL_REQUIRED: string = COMPANY_AUTH_PROTECTED_ROUTE + template.STILL_REQUIRED;
-export const CONFIRMATION: string = COMPANY_AUTH_PROTECTED_ROUTE + "confirmation";
```
ICoxCH: I'd argue to leave this line in over the new version, as the "confirmation" string only appears once (now that we're using string.replace() to create the url), not worth having a constant for "confirmation"

src/model/page.urls.ts
```diff
@@ -10,14 +10,18 @@ export const PROMISE_TO_FILE: string = SEPARATOR + "promise-to-file";
...
+export const CONFIRMATION: string = COMPANY_AUTH_PROTECTED_ROUTE + CONFIRMATION_APPEND;
+
 /**
  * URLs for redirects will need to start with the application name
  */
 export const PROMISE_TO_FILE_COMPANY_NUMBER: string = PROMISE_TO_FILE + COMPANY_NUMBER;
 export const PROMISE_TO_FILE_CHECK_COMPANY: string = PROMISE_TO_FILE + CHECK_COMPANY;
+export const PROMISE_TO_FILE_CONFIRMATION: string =
```
ICoxCH: Could this not just be PROMISE_TO_FILE_CONFIRMATION: string = PROMISE_TO_FILE + CONFIRMATION?
As above I believe the CONFIRMATON_APPEND constant is not required right now as the string "confirmation" only appears once

test/controllers/still.required.controllers.spec.unit.ts
```diff
@@ -34,4 +34,55 @@ describe("still required validation tests", () => {
...
+    toHaveBeenCalledWith(buildDummySession("123", "00006400"), IS_STILL_REQUIRED, false);
+  });
+
+  it("should update session when nO is selected", async () => {
+
+    const response = await request(app)
+        .post("/promise-to-file/company/00006400/still-required")
+        .set("Referer", "/")
+        .set("Cookie", [`${COOKIE_NAME}=123`])
+        .send({stillRequired: "nO"});
```
ICoxCH: Is n0 being interpreted as false here? Should we be doing that? I can understand making it case insensitive, but I personally wouldn't accept n0 as valid input here

mattch1: It is just to check that the to upper case is working properly it is a capital 'O' not a zero.

PR #58, opened 2020-02-14
Updated feedback link

SVarallo:
Also fixed still-required page validation

### DIFF
--- a/src/controllers/still.required.controller.ts
+++ b/src/controllers/still.required.controller.ts
@@ -8,7 +8,7 @@ import {getPromiseToFileSessionValue} from "../services/session.service";
 import {COMPANY_PROFILE} from "../session/keys";
 
 const validators = [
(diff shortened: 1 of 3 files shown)


src/controllers/still.required.controller.ts
```diff
@@ -8,7 +8,7 @@ import {getPromiseToFileSessionValue} from "../services/session.service";
 import {COMPANY_PROFILE} from "../session/keys";
 
 const validators = [
-  check("continuedIllness").not().isEmpty().withMessage(COMPANY_REQUIRED_NOT_SELECTED),
+  check("stillRequired").not().isEmpty().withMessage(COMPANY_REQUIRED_NOT_SELECTED),
```
mattch1: This has been done on ticket LFA-1380

SVarallo: Yes but this validation error is holding up the ticket that this bug corresponds to.

PR #59, opened 2020-02-17
Making template.paths an enum

markpit:


src/controllers/check.company.controller.ts
```diff
@@ -1,16 +1,16 @@
-import {NextFunction, Request, Response} from "express";
-import {PTFCompanyProfile} from "../model/company.profile";
-import {CHECK_COMPANY} from "../model/template.paths";
-import {getPromiseToFileSessionValue} from "../services/session.service";
-import {COMPANY_PROFILE} from "../session/keys";
+import { NextFunction, Request, Response } from "express";
+import { PTFCompanyProfile } from "../model/company.profile";
+import { Templates } from "../model/template.paths";
```
mwestacott: Presume that we could also import Templates.CHECK_COMPANY here? If yes, that would avoid repeating 'Templates.' twice further down and be consistent with the COMPANY_PROFILE import too.

src/model/template.paths.ts
```diff
@@ -1,6 +1,8 @@
-export const INDEX: string = "index";
-export const COMPANY_NUMBER: string = "company-number";
-export const CHECK_COMPANY: string = "check-company";
-export const STILL_REQUIRED: string = "still-required";
-export const CONFIRMATION_STILL_REQUIRED: string = "confirmation-still-required";
-export const CONFIRMATION_NOT_REQUIRED: string = "confirmation-not-required";
+export enum Templates {
```
mwestacott: Yes, like this. It's better.

src/controllers/company.number.controller.ts
```diff
@@ -7,7 +7,7 @@ import {COMPANY_NOT_FOUND, COMPANY_NUMBER_TOO_LONG,
     INVALID_COMPANY_NUMBER, NO_COMPANY_NUMBER_SUPPLIED} from "../model/error.messages";
 import {createGovUkErrorData, GovUkErrorData} from "../model/govuk.error.data";
 import {PROMISE_TO_FILE_CHECK_COMPANY} from "../model/page.urls";
-import {COMPANY_NUMBER} from "../model/template.paths";
+import {Templates} from "../model/template.paths";
```
SVarallo: So are we going do format these import like { this } or {this}? It should be consistent throughout the service

markpit: { xxxx } seems to be the standard https://www.typescriptlang.org/docs/handbook/modules.htmlhttps://www.typescriptlang.org/docs/handbook/modules.html

src/controllers/check.company.controller.ts
markpit: Wouldn't it read better if it were Templates.CHECK_COMPANY instead of CHECK_COMPANY?

mwestacott: I think in these cases it's fairly intuitive, but I don't mind. However, if we do stick with this (as a new "standard") shouldn't you refactor the COMPANY_PROFILE import?

markpit: this pr was just to do the templates. Not sure about session keys as it will mean changing all the session modules and the plan is to replace it when we get the common component.

markpit: Tried using import Templates.COMPANY_NUMBER but you can't import the individual type of the Enum, only the enum itself is exported.

PR #60, opened 2020-02-18
Feature/lfa 1403 api call

mattch1:
Promise to file api call.


src/session/config.ts
```diff
@@ -0,0 +1,12 @@
+const getEnvironmentValue = (key: string, defaultValue?: any): string => {
```
markpit: we already have a file for this - properties.ts

mattch1: Moved this.

src/client/axios.api.call.handler.ts
```diff
@@ -0,0 +1,30 @@
+import {AxiosError, AxiosRequestConfig, AxiosResponse, Method} from "axios";
+import axios from "axios";
+import logger from "../logger";
+
+export const HTTP_POST: Method = "post";
```
markpit: this module needs unit tests

mattch1: It has them.

src/client/apiclient.ts
```diff
@@ -1,9 +1,12 @@
+import {AxiosRequestConfig, AxiosResponse} from "axios";
```
mwestacott: Standard seems to be adding a space after { and before the } braces. You did this on line 7 so suggest doing here and on line 9.

src/controllers/confirmation.controller.ts
```diff
@@ -27,9 +29,19 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
   // TODO LFA-1169 Clarify what the ext deadline period will be for accounts and cs.
 
-  // TODO LFA-TBC call promise-to-file api
-
   const isStillRequired: boolean = getPromiseToFileSessionValue(req.chSession, IS_STILL_REQUIRED);
+
+  const token: string = req.chSession.accessToken() as string;
+  try {
+    if (token) {
+      // TODO  LFA-TBC Add isSubmitted flag to prevent this being sent twice
```
mwestacott: We should create the JIRA task at this point and reference it here, rather than use TBC. I can't actually see anything in the story we're doing about preventing duplicate submissions so does this need clarifying by an analyst?

src/client/apiclient.ts
markpit: I'd suggest leaving it for now as we should really change the everywhere. Maybe a tech debt story?

mwestacott: Consensus: do this later as a tidy up task we think. @markpit to create a tech task/story.

markpit: LFA-1405 created to cover the imports

src/controllers/confirmation.controller.ts
mattch1: Created ticket LFA-1406 and updated TODO comment.

PR #61, opened 2020-02-18
Fixing imports

markpit:
LFA-1405
Fixng imports syntax and also fixed a few minor things that caught my attention like adding javadoc, fixing indentation etc..


markpit:
> Is there a lint and sonar rule we can put in place to ensure that this is maintained?

Added config to ts-lint now so it will be maintained.

PR #72, opened 2020-02-28
Changed location of api to internalapi

ICoxCH:
API has moved to internal domain so updating location


ICoxCH:
https://github.com/companieshouse/chs-configs/pull/879

PR #73, opened 2020-03-03
stub screen to replace yes confirmation screen

ICoxCH:


src/controllers/confirmation.controller.ts
```diff
@@ -39,15 +39,18 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
 
   const token = req.chSession.accessToken() as string;
   try {
-      // TODO  LFA-1406 Add isSubmitted flag to prevent this being sent twice
+    // TODO  LFA-1406 Add isSubmitted flag to prevent this being sent twice
+    // TODO  LFA-1169: subtask Remove feature flag when stub screen removed
```
mwestacott: I would change comment to not mention 'feature flag' since this is confusing. Maybe just 'Remove check for still required when stub screen is removed'.

PR #81, opened 2020-03-10
Changes promise-to-file urls to company-required

SVarallo:


routes.yaml
```diff
@@ -3,4 +3,4 @@ group: web
 
 weight: 1000
 routes:
-  1: ^/promise-to-file/*
\ No newline at end of file
+  1: ^/company-required/*
```
mwestacott: The story says the URL should be 'beta.companieshouse.gov.uk/companyrequired' but this suggest it will be 'beta.companieshouse.gov.uk/company-required'.

SVarallo: Story has now been changed to have 'beta.companieshouse.gov.uk/company-required' as the url for the service

PR #82, opened 2020-03-12
LFA-1502 Get new deadline date from API and show on the page

mwestacott:


src/controllers/confirmation.controller.ts
```diff
@@ -39,33 +41,42 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
-        });
-    } else {
-      return res.render(Templates.COMPANY_REQUIRED,
-        {
-          company: companyProfile,
-          userEmail: email,
-        });
-    }
+    const newFilingDeadline = apiResponseData.new_filing_deadline;
+
```
markpit: I would check if newFilingDeadline is not null/undefined otherwise page will show with a blank date. In which case log error and return next(e) to show error page

mwestacott: Yep. Good idea.

mwestacott: Addressed this now.

PR #84, opened 2020-03-18
Added extra logging and added jquery back in for piwik

SVarallo:
Also did an npm audit fix which updated the nunjucks version

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "morgan": "~1.9.1",
     "msgpack5": "^4.2.1",
     "node-jose": "^1.1.3",
-    "nunjucks": "^3.1.7",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "morgan": "~1.9.1",
     "msgpack5": "^4.2.1",
     "node-jose": "^1.1.3",
-    "nunjucks": "^3.1.7",
+    "nunjucks": "^3.2.1",
```
ICoxCH: Have you done a brief sanity check on all the web pages with this new version?

SVarallo: Yep all seems to be working fine. Unit tests too.

PR #85, opened 2020-03-19
Amend still required confirmation screen fixed lint error

mattch1:
What happens next text amendment - updated
Text states new deadline – 28 days - already done
Late filing penalty  - already done

Fixed a lint error in a unit test -line too long.

### DIFF
--- a/views/confirmation-still-required.html
+++ b/views/confirmation-still-required.html
@@ -43,12 +43,19 @@ Confirmation page
         </p>
 
         <h2 class="govuk-heading-m">What happens next</h2>
-        <p>{{ company.companyName }}
-            will be kept on the register but it must file its accounts. The original filing deadline has not been changed or extended,
-            so the company will receive a
+        <p>
+            {{ company.companyName }} will be kept on the register. This means we will no longer close the company,
+            but it must file its accounts to stay on the register.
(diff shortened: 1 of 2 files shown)


views/confirmation-still-required.html
```diff
@@ -43,12 +43,19 @@ <h1 class="govuk-panel__title">You've told us
         </p>
 
         <h2 class="govuk-heading-m">What happens next</h2>
-        <p>{{ company.companyName }}
-            will be kept on the register but it must file its accounts. The original filing deadline has not been changed or extended,
-            so the company will receive a
+        <p>
+            {{ company.companyName }} will be kept on the register. This means we will no longer close the company,
```
mwestacott: Wondering why this text has changed from what I recently delivered in story LFA-1169 - the new text doesn't match what's on the screenshot of that story? No JIRA reference on the commit message so just wondering what requirement you're working off?

mwestacott: Ah, I've just seen the JIRA ref in the branch name. Is it a standard to have it at start of commit message too? Would be good to include it as useful when reading Git logs back at a later date and understanding why changes were made. Not enough to block the merge, so up to you.

PR #86, opened 2020-03-30
Feature/lfa 1406 application already submitted flag

mattch1:
As well as setting the submitted flag we also require access to the new deadline in cases where the application has and hasn't been submitted for the purpose of redisplaying the confirmation page.

### DIFF
--- a/src/controllers/confirmation.controller.ts
+++ b/src/controllers/confirmation.controller.ts
@@ -56,7 +56,6 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
     const token = req.chSession.accessToken() as string;
     let apiResponseData: any;
     try {
-        // TODO  LFA-1406 Add isSubmitted flag to prevent this being sent twice
 
         const axiosResponse: AxiosResponse = await callPromiseToFileAPI(companyProfile.companyNumber,
             token, isStillRequired);
@@ -77,7 +76,6 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
 
   if (isStillRequired) {
     const filingDueOn =  getPromiseToFileSessionValue(req.chSession, NEW_DEADLINE);
-    logger.debug("\n\n>>>>>>>>>>>>>>>>>>>> NEW DEADLINE " + filingDueOn + "\n\n");
     logger.debug(`New filing deadline : ${filingDueOn}`);
 
     if (!filingDueOn) {
--- a/src/controllers/confirmation.controller.ts
+++ b/src/controllers/confirmation.controller.ts
@@ -57,25 +57,25 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
     let apiResponseData: any;
     try {
 
-        const axiosResponse: AxiosResponse = await callPromiseToFileAPI(companyProfile.companyNumber,
-            token, isStillRequired);
+      const axiosResponse: AxiosResponse = await callPromiseToFileAPI(companyProfile.companyNumber,
+        token, isStillRequired);
 
-        apiResponseData = axiosResponse.data;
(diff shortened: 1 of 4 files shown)


src/controllers/confirmation.controller.ts
```diff
@@ -51,25 +51,31 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
-    // TODO  LFA-1406 Add isSubmitted flag to prevent this being sent twice
+  const isSubmitted: boolean = getPromiseToFileSessionValue(req.chSession, ALREADY_SUBMITTED);
+  if (!isSubmitted) {
+    const token = req.chSession.accessToken() as string;
+    let apiResponseData: any;
+    try {
 
-    const axiosResponse: AxiosResponse = await callPromiseToFileAPI(companyProfile.companyNumber,
-        token, isStillRequired);
+        const axiosResponse: AxiosResponse = await callPromiseToFileAPI(companyProfile.companyNumber,
```
mwestacott: Looks like you've switched to 4 spaces for a tab, but 2 is the standard.

src/controllers/confirmation.controller.ts
```diff
@@ -51,25 +51,31 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
+            token, isStillRequired);
 
-    apiResponseData = axiosResponse.data;
+        apiResponseData = axiosResponse.data;
+        await updatePromiseToFileSessionValue(req.chSession, NEW_DEADLINE, apiResponseData.filing_due_on);
```
mwestacott: I don't understand why this is being done. It isn't mentioned in the requirements/story and why does it need to be stored in the session?

src/controllers/confirmation.controller.ts
```diff
@@ -51,25 +51,31 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
+        return next(e);
+    }
+  } else {
+      logger.error("Form already submitted, not processing again");
   }
 
   if (isStillRequired) {
-    const filingDueOn = apiResponseData.filing_due_on;
-
+    const filingDueOn =  getPromiseToFileSessionValue(req.chSession, NEW_DEADLINE);
```
mwestacott: Two spaces to right of equals sign. Should/can npm lint be picking up these formatting issues?

src/controllers/confirmation.controller.ts
mattch1: As stated above even when submitted is set to true the code still needs the new deadline so that it can be displayed on the confirmation screen the only thing that should be different is the api client is not called. Therefore we need somewhere global to access the new deadline data.

src/controllers/confirmation.controller.ts
mattch1: Changed to 2 spaces.

src/controllers/confirmation.controller.ts
mattch1: Removed extra space.

test/controllers/confirmation.controller.spec.unit.ts
```diff
@@ -139,13 +140,15 @@ describe("Company no longer required confirmation screen tests", () => {
...
+    mockPTFSession.mockImplementationOnce(() => false);
+    mockPTFSession.mockImplementationOnce(() => "2028-03-10");
 
     mockCallProcessorApi.prototype.constructor.mockImplementationOnce(() => Promise.resolve((
-      {
-        data: {
-          filing_due_on: "2028-03-10",
-        },
-      } )));
+        {
```
mwestacott, line 147: Extra 2 spaces in tab that have been added should go.

mattch1:
Closing this PR and the branch - as there is a simpler solution to this.

PR #87, opened 2020-04-03
Create not eligible screen and show when appropriate

SVarallo:


src/controllers/confirmation.controller.ts
```diff
@@ -70,12 +72,21 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
   if (isStillRequired) {
     const filingDueOn = apiResponseData.filing_due_on;
 
+    if (apiResponseStatus === 400) {
+      const cannotUseReason: string = eligibilityReasonCode[apiResponseData.reason_code];
```
ICoxCH, line 76: If somehow apiResponseData.reason_code is null. What happens here (Just thinking if the api returns a 400 for some reason other than the eligibility check fails)

mwestacott:
I can't remember - should/do changes to the package-lock.json file need to be committed?

src/controllers/confirmation.controller.ts
mwestacott: Spec says that for a 400 response reason_code should always be set, so if it isn't we should probably just show the 'Sorry, service failed' screen I guess?

test/controllers/confirmation.controller.spec.unit.ts
```diff
@@ -163,6 +164,154 @@ describe("Company no longer required confirmation screen tests", () => {
...
+    loadCompanyAuthenticatedSession(mockCacheService, COMPANY_NUMBER, EMAIL);
+    const dummyProfile = getDummyCompanyProfile(true, true);
+    mockPTFSession.mockImplementationOnce(() => dummyProfile);
+    mockPTFSession.mockImplementationOnce(() => true);
+    mockActiveFeature.mockImplementationOnce(() => true);
+
+    mockCallProcessorApi.prototype.constructor.mockImplementation(() => Promise.resolve((
+      {
+        data: {
+          filing_due_on: "null",
```
mwestacott, line 183: According to the spec, filing_due_on field isn't even present in the JSON for a 400 response. So I'm interested in what the API does actually return? Did/could you try removing the "null" value here (and does test still pass)?

src/model/eligibilityReasonCode.ts
```diff
@@ -0,0 +1,7 @@
+export enum eligibilityReasonCode {
+  COMPANY_IS_ELIGIBLE = "company is eligible",
```
mwestacott, line 2: So these values don't seem to match the current return codes from the API - @mattch1 did merge a PR recently which changed these. So I think they need to correspond to what's in the API **_EligibilityReturnCodeToWeb_** Java enum.

SVarallo: Yep, I'll change the values to match those in the api. The strings okay to stay as they are?

test/controllers/confirmation.controller.spec.unit.ts
SVarallo: They pass without the filing_due_on field. I'll remove them 👍

SVarallo:
> I can't remember - should/do changes to the package-lock.json file need to be committed?

Yes, changes to the package-lock do need to be committed 👍

src/model/eligibilityReasonCode.ts
mwestacott: 🆗 Yes, string values are fine, assuming they're still sensible for the values in EligibilityReturnCodeToWeb.

PR #88, opened 2020-04-07
Feature/lfa 1612 allow single letter companies

mattch1:
Allow companies with single letetr prefix to access company details - changed the regex to allow 1 or 2.

Changed padding to check for two letter prefixes first (as two letters are considered to be 1 lettered as well by regex) if not code checks for 1 letetr and pads.

Unit tests for new regex validation and for the padding.

### DIFF
--- a/src/controllers/company.number.controller.ts
+++ b/src/controllers/company.number.controller.ts
@@ -36,7 +36,6 @@ const formatCompanyNumber = (companyNumber: string, leadPoint: number, padStart:
   const leadingLetters = companyNumber.substring(0, leadPoint);
   let trailingChars = companyNumber.substring(leadPoint, companyNumber.length);
   trailingChars = trailingChars.padStart(padStart, "0");
-  logger.debug("`n >>>>>>>>>>> TRAILING CHARS: " + trailingChars);
   return leadingLetters + trailingChars;
 };
 
--- a/src/controllers/company.number.controller.ts
+++ b/src/controllers/company.number.controller.ts
@@ -21,11 +21,10 @@ const preValidators = [
 // pads company number to 8 digits with 0's and removes whitespace
 const padCompanyNumber = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
   let companyNumber: string = req.body.companyNumber;
-  if (/^([a-zA-Z]{2}?)/gm.test(companyNumber)) {
-    const leadingLetters = companyNumber.substring(0, 2);
-    let trailingChars = companyNumber.substring(2, companyNumber.length);
-    trailingChars = trailingChars.padStart(6, "0");
-    companyNumber = leadingLetters + trailingChars;
+  if (/^([a-zA-Z]{1}?)/gm.test(companyNumber)) {
+    companyNumber = formatCompanyNumber(companyNumber, 1, 7);
+  } else if (/^([a-zA-Z]{2}?)/gm.test(companyNumber)) {
+    companyNumber = formatCompanyNumber(companyNumber, 2, 6);
   } else {
     companyNumber = companyNumber.padStart(8, "0");
   }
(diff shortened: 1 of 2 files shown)


src/controllers/company.number.controller.ts
```diff
@@ -22,21 +22,27 @@ const preValidators = [
...
+  let trailingChars = companyNumber.substring(leadPoint, companyNumber.length);
+  trailingChars = trailingChars.padStart(padStart, "0");
+  return leadingLetters + trailingChars;
+};
+
 // validator middleware that checks for invalid characters in the input
 const postValidators = [
   check("companyNumber").blacklist(" ").escape().custom((value: string) => {
-    if (!/^([a-zA-Z]{2})?[0-9]{6,8}$/gm.test(value)) {
+    if (!/^([a-zA-Z]{1,2})?[0-9]{6,8}$/gm.test(value)) {
```
ICoxCH, line 45: Does this validator work if a 1 letter, 7 number string was passed to it?
Looks to me as if its expecting 'one or none' of 1 letter, or 2 letters.
Followed by either 6 numbers, or 8 numbers.
(My regex is rusty so I may be missing something here)

markpit: It allows a 1 letter 6 number through which I don't think is correct.
H123456
you can test it on https://regex101.com/

src/controllers/company.number.controller.ts
```diff
@@ -22,21 +22,27 @@ const preValidators = [
 const padCompanyNumber = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
   let companyNumber: string = req.body.companyNumber;
   if (/^([a-zA-Z]{2}?)/gm.test(companyNumber)) {
-    const leadingLetters = companyNumber.substring(0, 2);
-    let trailingChars = companyNumber.substring(2, companyNumber.length);
-    trailingChars = trailingChars.padStart(6, "0");
-    companyNumber = leadingLetters + trailingChars;
+    companyNumber = formatCompanyNumber(companyNumber, 2, 6);
```
mwestacott, line 25: If we know a company number is always 8 chars long (I think it is, and the two scenarios here suggest that), could you dispense with the second parameter to your 'formatCompanyNumber' function and just calculate the 'padStart' variable as '8 - leadPoint'?

src/controllers/company.number.controller.ts
ICoxCH: Seems it also lets through 'just 6 numbers' and '2 letters 8 numbers'

markpit: try this : ^[0-9]{8}|([a-zA-Z]{1})[0-9]{7}|([a-zA-Z]{2})[0-9]{6}$

mattch1: This is based on existing pattern in extensions: 
if (!/^([a-zA-Z]{2})?[0-9]{6,8}$/gm.test(value)) {

so I thought we were aware of its shortcomings as this allows thing like:
HA12345678
345678

If it is of concern then we need to raise a ticket to fix this in extensions as well.

ICoxCH: ^[0-9]{8}$|^([a-zA-Z]{1})[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$
This appears to work? Adds start and end of string checks to each or section

mattch1: I understand why this ipassed previously if we look at the code as a whole the regex check happens in the post validator after we pad shart code such as "H123456" with 0s so this becomes H0123456 there is also a check to ensure that the string is 8 characters long.

src/controllers/company.number.controller.ts
mattch1: Updated.

src/controllers/company.number.controller.ts
ICoxCH: Should we update the validator anyway? I get that currently it all works as expected.
But if in future the padding method changes or doesn't work as expected. We want to ensure the validator would pick that up and fail it?

markpit: can we update this regex to ^[0-9]{8}$|^([a-zA-Z]{1})[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$. to make it more accurate

mattch1: The unit tests would alert us to that - there is test that looks at 6400 and expects 00006400, this would start to fail and prompt further changes. However, we want the code to be as robust as it can be so will change.

mattch1: Updated regex to this new pattern.

PR #96, opened 2020-04-15
Updated jest, kind-of and nodemon to fix vulnerabilities

SVarallo:
Jira ticket at: https://companieshouse.atlassian.net/browse/LFA-1633


package.json
```diff
@@ -37,11 +37,12 @@
     "@types/jest": "^24.0.0",
     "@types/nunjucks": "^3.1.1",
     "@types/supertest": "^2.0.7",
-    "jest": "^24.1.0",
+    "jest": "^25.3.0",
```
ICoxCH, line 40: i believe ts-jest should be updated to match this version

SVarallo: Updated to 25.3.1 👍

PR #97, opened 2020-04-16
LFA-1628 Report confirmation statement as late when appropriate

mwestacott:


src/controllers/confirmation.controller.ts
```diff
@@ -92,10 +92,21 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
       return next(new Error("No new filing due date returned by the PTF API"));
     }
 
+    let overdueFiling: string;
+    if (companyProfile.isAccountsOverdue && !companyProfile.isConfirmationStatementOverdue) {
+      overdueFiling = "accounts";
+    } else if (companyProfile.isAccountsOverdue && companyProfile.isConfirmationStatementOverdue) {
+      overdueFiling = "accounts and confirmation statement";
+    } else {
+      // TODO Handle scenario when only confirmation statement is overdue LFA-1479
```
mwestacott, line 101: Tackling this story next...

src/controllers/confirmation.controller.ts
```diff
@@ -92,10 +92,21 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
       return next(new Error("No new filing due date returned by the PTF API"));
     }
 
+    let overdueFiling: string;
+    if (companyProfile.isAccountsOverdue && !companyProfile.isConfirmationStatementOverdue) {
```
markpit, line 96: how about:
`let overdueFiling: string;
    if (companyProfile.isAccountsOverdue) {
      overdueFiling = "accounts";
      if (companyProfile.isConfirmationStatementOverdue) {
        overdueFiling = overdueFiling.concat(" and confirmation statement");
      }
    } else {
      // TODO Handle scenario when only confirmation statement is overdue LFA-1479
      overdueFiling = "UNDEFINED";
    }`

src/controllers/confirmation.controller.ts
```diff
@@ -92,10 +92,21 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
       return next(new Error("No new filing due date returned by the PTF API"));
     }
 
+    let overdueFiling: string;
```
markpit, line 95: how about:
let overdueFiling: string;
    if (companyProfile.isAccountsOverdue) {
      overdueFiling = "accounts";
      if (companyProfile.isConfirmationStatementOverdue) {
        overdueFiling = overdueFiling.concat(" and confirmation statement");
      }
    } else {
      // TODO Handle scenario when only confirmation statement is overdue LFA-1479
      overdueFiling = "UNDEFINED";
    }

src/controllers/confirmation.controller.ts
markpit, line 96:
```suggestion
    let overdueFiling: string;
    if (companyProfile.isAccountsOverdue) {
      overdueFiling = "accounts";
      if (companyProfile.isConfirmationStatementOverdue) {
        overdueFiling = overdueFiling.concat(" and confirmation statement");
      }
    } else {
      // TODO Handle scenario when only confirmation statement is overdue LFA-1479
      overdueFiling = "UNDEFINED";
    }
```

markpit: what do you think?

markpit: actually forget it, the conf statement bit would need another if, whereas yours doesn't

PR #98, opened 2020-04-17
LFA-1479 Show correct text when only confirmation statement is overdue

mwestacott:


src/controllers/confirmation.controller.ts
```diff
@@ -119,4 +115,18 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
   }
 };
 
+const getOverdueFiling = (companyProfile: PTFCompanyProfile): string => {
+  let overdueFiling: string = "";
+
+  if (companyProfile.isAccountsOverdue && !companyProfile.isConfirmationStatementOverdue) {
```
mwestacott, line 121: I think this is the most readable format. @markpit what do you think?

markpit: looks good

markpit: there is actually a trick I discovered.
const getOverdueFiling = ({isAccountsOverdue, isConfirmationStatementOverdue}): string =>

call the method as normal, you can then use the fields from companyProfile directly!

mwestacott: Yes, that seems to work a treat. Updated.

PR #100, opened 2020-04-21
back link changes for warning screen

mattch1:
Back links to all screens except the two authentication screens:
login
authentication code

A typical back link journey would be:
still-required -> warning -> check-company -> company-number -> company-required (start screen)

Changes here are to add the warning screen to the back link journey


src/controllers/still.required.controller.ts
```diff
@@ -22,9 +22,12 @@ const validators = [
  */
 export const getRoute = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 
-  const companyName: string = getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE).companyName;
+  const company: PTFCompanyProfile = getPromiseToFileSessionValue(req.chSession, COMPANY_PROFILE);
+  const companyName: string = company.companyName;
+  const backLinkUrl: string = `/company-required/company/${company.companyNumber}/warning`;
```
markpit, line 27: can this go i the Templates file? to avoid hard coding urls

markpit: or just keep the url in the html and just pass the companyNumber into the html

markpit: further down in this same file on line 56 it uses the pageUrls, you should do the same

const url = COMPANY_REQUIRED_CONFIRMATION.replace(":companyNumber", companyProfile.companyNumber);

mattch1: Removed hardcoding from the code - added this to the url file and as suggested replace :companyNumber.

PR #101, opened 2020-04-21
Added print link to confirmation-still-required page

SVarallo:
Jira ticket at https://companieshouse.atlassian.net/browse/LFA-1536

This will open up a print window instead of a whole new page, with the required stylings.

### DIFF
--- a/views/confirmation-still-required.html
+++ b/views/confirmation-still-required.html
@@ -38,8 +38,8 @@ Confirmation page
         </p>
         {% endif %}
 
-        <p class="govuk-body">
-            <a href="print-confirmation" class="govuk-link">View printer-friendly version</a>
+        <p class="govuk-body hide-print-content print-link no-print" style="display: block;" aria-hidden="true" hidden="hidden">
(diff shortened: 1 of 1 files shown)


views/confirmation-still-required.html
```diff
@@ -38,8 +38,8 @@ <h1 class="govuk-panel__title">You've told us
         </p>
         {% endif %}
 
-        <p class="govuk-body">
-            <a href="print-confirmation" class="govuk-link">View printer-friendly version</a>
+        <p class="govuk-body hide-print-content print-link no-print" style="display: block;" aria-hidden="true" hidden="hidden">
+            <a href="javascript:if(window.print)window.print()" class="govuk-link">View printer-friendly version</a>
```
markpit, line 42: what happens if JS is disabled?

SVarallo: Then the link will be hidden and they will have to do it the usual way. They should have an email copy anyway. Same with screen readers.

markpit: ok

PR #104, opened 2020-04-23
Feature/lfa 1406 prevent duplicate submits

mattch1:
Flag to prevent confirmation screen back button click to the still required page so the use cannot repeat submits when company is still required. This is not the case for comapnies that are no longer required.

If a user tries to resubmit a still required application then on clicking the back button they are redirected to a custom screen stating that they cannot use this service as the application has already been received.

Unit tests for the routing the new screen and to check that the session update for the flag is called on confirmation for still required and not for no longer required applications.


markpit:
I've raised a potential issue in the Jira story, If they select No, go back and select No again, we'll get duplicate chips contacts created and duplicate emails sent.

PR #107, opened 2020-04-28
Amended confirmation not required screen and added unit tests.

SVarallo:
Jira tickets at: https://companieshouse.atlassian.net/browse/LFA-1530
                        https://companieshouse.atlassian.net/browse/LFA-1531
                        https://companieshouse.atlassian.net/browse/LFA-1532

Additional unit tests are to accommodate for the variations in the confirmation screen. Also split up the tests for the different confirmation pages.

Updated mock utils getDummyCompanyProfile function to allow setting the confirmation overdue variable separately for more specific screen testing.

### DIFF
--- a/src/controllers/confirmation.controller.ts
+++ b/src/controllers/confirmation.controller.ts
@@ -69,6 +69,7 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
     return next(e);
   }
 
+  const overdueFiling: string = getOverdueFiling(companyProfile);
   if (isStillRequired) {
     const filingDueOn = apiResponseData.filing_due_on;
 
@@ -92,12 +93,6 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
       return next(new Error("No new filing due date returned by the PTF API"));
     }
 
-    const overdueFiling: string = getOverdueFiling(companyProfile);
-
-    if (!overdueFiling) {
-      return next(new Error("Company still required but neither accounts or confirmation statement are overdue"));
-    }
-
     return res.render(Templates.CONFIRMATION_STILL_REQUIRED,
       {
         company: companyProfile,
@@ -109,7 +104,8 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
     return res.render(Templates.CONFIRMATION_NOT_REQUIRED,
       {
         company: companyProfile,
-        overdueFiling: getOverdueFiling(companyProfile),
+        overdueFiling,
+        showLFPWarning: companyProfile.isAccountsOverdue,
         userEmail: email,
       });
   }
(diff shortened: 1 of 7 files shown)


test/controllers/confirmation.controller.spec.unit.ts
```diff
@@ -128,6 +200,14 @@ describe("Company no longer required confirmation screen tests", () => {
     expect(resp.text).toContain("BT2 8BG");
     expect(resp.text).toContain(PAGE_TITLE);
   });
+});
+
+describe("Company still required confirmation screen tests", () => {
+
+  const mockCacheService = loadSession as jest.Mock;
```
markpit, line 207: you can move these 4 mocks so they are declared before the first describe(), you won't have to repeat them then

markpit: and delete the others from the other describe()

SVarallo: Okay, will do.

views/confirmation-not-required.html
```diff
@@ -24,16 +28,17 @@ <h1 class="govuk-panel__title">You've told us
...
         </p>
 
-        <p>
-            {{company.companyName}}
-            will receive a
-            <a href="https://www.gov.uk/annual-accounts/penalties-for-late-filing">late filing penalty</a>
-            that increases over time until the company is closed.</p>
+        <p>If you do not close the company and do not file your overdue {{ overdueFiling }}, we may prosecute the company directors.</p>
+
+        {% if (overdueFiling === 'accounts') or (overdueFiling === 'accounts and confirmation statement') %}
```
markpit, line 36: it would be better to pass in a boolean flag here to indicate if this text block needs to be shown. We should keep as much decision making out of the templates, makes it easier to maintain in the future. In the controller you could pass into the template , showLFPWarning: company.isAccountsOverdue

SVarallo: Done

src/controllers/confirmation.controller.ts
```diff
@@ -124,6 +120,8 @@ const getOverdueFiling = ({isAccountsOverdue, isConfirmationStatementOverdue}):
     overdueFiling = "confirmation statement";
   } else if (isAccountsOverdue && isConfirmationStatementOverdue) {
     overdueFiling = "accounts and confirmation statement";
+  } else if (!isAccountsOverdue && !isConfirmationStatementOverdue) {
+    overdueFiling = "no open compliance case";
```
mwestacott, line 124: overdueFiling no longer seems to be the right name to use for this variable and function name as 'no open compliance case' is not an overdue filing. Is it now just holding a general message? Couldn't see in the template changes below where it's used, so not sure...

SVarallo: At the moment there is no handling for the scenario of when there's nothing overdue and the customer presses "no". This string isn't required at the moment but it's there just until a new story is drafted to cover that scenario. Otherwise it would just return the screen with an empty overdueFiling. If you think it should be removed entirely that's totally fair and I'll do that however, because I agree it doesn't really serve any purpose.

mwestacott: Ok. I would just suggest maybe getting something like 'TBC' (To Be Completed) into the string that's output, then nice and clear that this is still a work in progress. I'd also add a comment to the code describing the above.

markpit: a // TODO would be better, if we put 'TBC' into the string we'll forget about it!

mwestacott: Agreed. I mean doing both. A // TODO comment and make that output text a bit clearer that it's really just a placeholder text for future functionality.

PR #108, opened 2020-04-30
LFA-1668 Increased unit-test coverage to 100% on confirmation controller

mwestacott:


src/controllers/confirmation.controller.ts
```diff
@@ -79,13 +79,12 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
...
         return next(new Error("No reason_code in api response"));
       }
-      if (cannotUseReason !== eligibilityReasonCode.COMPANY_IS_ELIGIBLE) {
-        return res.render(Templates.NOT_ELIGIBLE,
-          {
-            cannotUseReason,
-            companyName: companyProfile.companyName,
-          });
-      }
+
```
mwestacott, line 82: Removed unneeded conditional - COMPANY_IS_ELIGIBLE is never returned by API in the case of a 400 response. Have tested the app and all working without it.

src/controllers/confirmation.controller.ts
```diff
@@ -120,8 +119,9 @@ const getOverdueFiling = ({isAccountsOverdue, isConfirmationStatementOverdue}):
     overdueFiling = "confirmation statement";
   } else if (isAccountsOverdue && isConfirmationStatementOverdue) {
     overdueFiling = "accounts and confirmation statement";
-  } else if (!isAccountsOverdue && !isConfirmationStatementOverdue) {
-    // TODO handle this output with appropriate render when story is created.
+  } else {
```
mwestacott, line 122: The else logic was redundant since this was the only remaining scenario (plus it was preventing 100% coverage) so removed.

PR #109, opened 2020-04-30
Lfa 1668 company number controller test coverage

mwestacott:


src/controllers/company.number.controller.ts
```diff
@@ -72,15 +72,15 @@ const route = async (req: Request, res: Response, next: NextFunction): Promise<v
   const companyNumber: string = req.body.companyNumber;
 
   try {
-    logger.info("Retrieving company profile for company number ${companyNumber}");
+    logger.info(`Retrieving company profile for company number ${companyNumber}`);
```
mwestacott, line 75: Noticed that company number wasn't being output to logs here and below, so fixed that.

PR #111, opened 2020-05-01
Add company already in prosecution option to not eligible page

SVarallo:
Jira ticket at https://companieshouse.atlassian.net/browse/LFA-1637

### DIFF
--- a/views/not-eligible.html
+++ b/views/not-eligible.html
@@ -50,11 +50,13 @@ You cannot use this service
       <p>You do not need to let us know that the company is still required.</p>
     {% elseif cannotUseReason === 'company already in prosecution' %}
       <p>The company {{companyName}} is already late filing their {{overdueFiling}}. The company must now
-      <a href="https://www.gov.uk/running-a-limited-company">file its overdue
-        {{overdueFiling}}</a>.</p>
+        <a href="https://www.gov.uk/running-a-limited-company">file its overdue
+        {{overdueFiling}}</a>.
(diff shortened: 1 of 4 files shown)


views/not-eligible.html
```diff
@@ -48,13 +48,32 @@ <h1 class="govuk-heading-xl">You cannot use this service</h1>
     {% if cannotUseReason === 'no open compliance case' %}
       <p>The accounts and confirmation statement for {{companyName}} have been filed and are up to date. This service is for companies with overdue filings.</p>
       <p>You do not need to let us know that the company is still required.</p>
+    {% elseif cannotUseReason === 'company already in prosecution' %}
+      <p>The company {{companyName}} is already late filing their {{overdueFiling}}. The company must now
+        <a href="https://www.gov.uk/running-a-limited-company">file its overdue
+        {{overdueFiling}}</a>.
+      </p>
+      <p>We will send a letter to the directors of the company soon to warn them that the company could be prosecuted if the
+        {{overdueFiling}}
```
mwestacott, line 57: Looks like the English might read slightly wrong if only CS is late - "We will send a letter to the directors of the company soon to warn them that the company could be prosecuted if the confirmation statement are not filed."

Is this known/acceptable?

SVarallo: Added a variable in place of that word so if the overdueFiling equals "confirmation statement" it will become "is", otherwise, the word will be "are".

PR #112, opened 2020-05-01
LFA-1411 service unavailable screen

markpit:
Adding a service unavailable screen that will show if the config SHOW_SERVICE_UNAVAILABLE_PAGE=on

### DIFF
--- a/test/availability/middleware/service.availability.spec.unit.ts
+++ b/test/availability/middleware/service.availability.spec.unit.ts
@@ -45,6 +45,15 @@ describe("Availability tests", () => {
     expect(response.text).toContain(UNAVAILABLE_TEXT);
   });
 
+  it("should show the service unavailable page for non start page with referer", async () => {
+    process.env.SHOW_SERVICE_UNAVAILABLE_PAGE = "on";
+
+    const response = await request(app)
+      .get("/company-required/company/00006400/still-required")
+      .set("Referer", "/company-required/company/00006400/warning");
+    expect(response.text).toContain(UNAVAILABLE_TEXT);
+  });
+
   it("should NOT show the service unavailable page", async () => {
     process.env.SHOW_SERVICE_UNAVAILABLE_PAGE = "off";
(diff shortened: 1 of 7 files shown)


test/availability/middleware/service.availability.spec.unit.ts
```diff
@@ -0,0 +1,57 @@
...
+    const response = await request(app)
+      .get("/company-required/");
+    expect(response.text).toContain(UNAVAILABLE_TEXT);
+  });
+
+  it("should show the service unavailable page for non start page", async () => {
+    process.env.SHOW_SERVICE_UNAVAILABLE_PAGE = "on";
+
+    const response = await request(app)
+      .get("/company-required/company-number");
```
ICoxCH, line 44: this test may only be passing due to you not setting a referrer.
The lack or referer causes a redirect to the start page that then redirects to the unavailable page.
If a user is mid journey when we turn this flag on, do we want to allow them to finish the journey or stop them right away?

mwestacott: PR looks fine to me apart from this comment. I guess we'd want to stop all journeys immediately, no? If session timeout was/is 20 mins or something then you could have people completing journeys quite a bit later when others still couldn't get in.

markpit: yes we want to stop them right away. I'll look at adding another test with a referer in to simulate a user who's already in the app.

ICoxCH: If we'd like to stop them right away then in the app.ts we could just do 
`app.use(checkServiceAvailability)` instead of `app.use(`${pageURLs.COMPANY_REQUIRED}`, checkServiceAvailability);`

That way the middleware will run on all url's

markpit: code updated to check all urls and added extra test with referer

PR #118, opened 2020-05-14
LFA-1377 Clear PTF session when error screen is triggered

mwestacott:


src/app.ts
```diff
@@ -72,6 +73,10 @@ app.use((err, req, res, next) => {
   res.locals.message = err.message;
   res.locals.error = req.app.get("env") === "development" ? err : { };
 
+  // Any old PTF session data will be discarded as a result of this call to create a new PTF session object. The
+  // new, empty PTF session will be utilised if the user decides to start a new PTF journey
+  createPromiseToFileSession(req.chSession);
```
markpit, line 78: i think we may need to make this error function async as the createPromiseToFIleSession returns a promise. We could end up executing past the createPromiseToFileSession call before the promise returns - try call it with await and mark function as async.

PR #124, opened 2020-05-29
LFA-1765 Ensure missing text appears when only conf statement is late

mwestacott:


views/confirmation-still-required.html
```diff
@@ -48,6 +48,15 @@ <h2 class="govuk-heading-m">What happens next</h2>
             but it must file its {{ overdueFiling }} to stay on the register.
         </p>
 
+        <!-- This block is displayed if ONLY the confirmation statement is late: -->
```
ICoxCH, line 51: Would these comments be visible on the source for the actual page? May be better to put the comments in nunjucks so they aren't sent to the user
https://mozilla.github.io/nunjucks/templating.html#comments

mwestacott: Yes, visible in the source so will do as suggested.

views/confirmation-still-required.html
```diff
@@ -48,6 +48,15 @@ <h2 class="govuk-heading-m">What happens next</h2>
             but it must file its {{ overdueFiling }} to stay on the register.
         </p>
 
+        {# This block is displayed if ONLY the confirmation statement is late: #}
+        {% if (company.isAccountsOverdue === false) and (company.isConfirmationStatementOverdue === true) %}
+            <p>
+                If {{ company.companyName }} has not filed by {{ newDeadline }}, we may prosecute the directors of the company.
```
SVarallo, line 54: Bit picky I know but why is this a "new deadline" if it's not been changed or extended?

PR #143, opened 2021-05-14
Updated hosted-git-info to 2.8.9 to fix security issues

gevans-ch:
Resolves BI-7016: https://companieshouse.atlassian.net/browse/BI-7016


package.json
```diff
@@ -21,6 +21,7 @@
     "express": "~4.16.1",
     "express-validator": "^5.3.1",
     "govuk-frontend": "^3.11.0",
+    "hosted-git-info": "^2.8.9",
```
YiannisPapastergiou, line 24: If it was registered as a dev dependency I believe you should install it as a dev dependency now as well.

I can see this from opening the package-lock.joson file at line 3990

YiannisPapastergiou (approved):

PR #158, opened 2023-08-25
upgrade tslint to eslint

wdhda-ch:


okoye-companieshouse:
Hey @wdhda-ch is it possible to break this PR down? As it is not just upgrading from tslint to eslint?

okoye-companieshouse (approved):

PR #159, opened 2023-08-30
upgrade service to node18

wdhda-ch:


ecs-image-build/docker_start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Start script for promise-to-file-web
+
+PORT=3000
+
+export NODE_PORT=${PORT}
+exec node /opt/dist/App.js -- ${PORT}
```
amartin7663, line 8: This should be www.js, similar to here: https://github.com/companieshouse/confirmation-statement-web/blob/main/ecs-image-build/docker_start.sh

src/authentication/company/middleware/index.ts
```diff
@@ -2,7 +2,7 @@ import { NextFunction, Request, Response } from "express";
 import logger from "../../../logger";
 import {
   COOKIE_NAME, OAUTH2_AUTH_URI, OAUTH2_CLIENT_ID,
-  OAUTH2_REDIRECT_URI
+  OAUTH2_REDIRECT_URI,
```
amartin7663, line 5: This is the last property import, so shouldn't have a comma?

terraform/groups/ecs-service/locals.tf
```diff
@@ -0,0 +1,85 @@
...
+locals {
+  stack_name                = "company-requests" # this must match the stack name the service deploys into
+  name_prefix               = "${local.stack_name}-${var.environment}"
+  service_name              = "promise-to-file-web"
+  container_port            = "3000" # default node port required here until prod docker container is built allowing port change via env var
+  docker_repo               = "promise-to-file-web"
+  lb_listener_rule_priority = 13
+  lb_listener_paths         = ["^/company-required/*"]
+  healthcheck_path          = "/company-required" #healthcheck path for promise to file web
+  healthcheck_matcher       = "200-302"           # no explicit healthcheck in this service yet, change this when added!
```
amartin7663, line 11: This is incorrect now, should be the new healthcheck path, and the matcher can be 200 only

terraform/groups/ecs-service/profiles/development-eu-west-2/cidev/vars
```diff
@@ -0,0 +1,32 @@
...
+# oauth2_client_id
+
+account_web_url = "https://account.cidev.aws.chdev.org"
+
+api_url="https://api.cidev.aws.chdev.org"
+cdn_host = "d3uvya5a8a1ncx.cloudfront.net"
+chs_url = "https://cidev.aws.chdev.org"
+company_still_required_feature_flag=true
+cookie_domain = "cidev.aws.chdev.org"
+internal_api_url="http://api.chs-dev.internal"
```
amartin7663, line 28: This is in the secret config list, plus it's incorrect

